PR #3, opened 2019-04-24
feat: add name, email, & dob

abutterworth:
- No live data yet
- Includes redux pipes and stub functions for api calls


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -0,0 +1,161 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { injectIntl, intlShape } from 'react-intl';
+
+import messages from './AccountSettingsPage.messages';
+
+// Actions
```
albemarle: Are these comments necessary?  It seems pretty obvious from the code.  (Plus selectors doesn't get a comment?  Not that I think it should.  :-) )

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+};
+
+AccountSettingsPage.defaultProps = {
+  loading: false,
+  loaded: false,
+  loadingError: null,
+  openFormId: null,
+  fieldSections: [
+    {
+      sectionHeading: 'Account Information',
```
albemarle: i18n?

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+        },
+        {
+          name: 'email',
+          label: 'Email address',
+          type: 'email',
+          helpText: 'You receive messages from Your Platform Name Here and course teams at this address.',
+        },
+        {
+          name: 'yearofbirth',
+          label: 'Year of birth',
```
albemarle: Just a note when we do messages for real, Full Name is caps for each word, so probably Year of birth should be also.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -0,0 +1,161 @@
...
+          type: 'select',
+          options: (() => {
+            const currentYear = new Date().getFullYear();
+            const years = [];
+            let startYear = currentYear - 120;
```
albemarle: Optimistic, I like it.

src/account-settings/components/temp/Input.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  getHTMLTagForType() {
+    const { type } = this.props;
+    if (type === 'select' || type === 'textarea') return type;
+    return 'input';
+  }
+
+  getClassNameForType() {
+    const { type } = this.props;
+    if (type === 'file') return 'form-control-file';
```
albemarle: Might be nicer as a switch?

src/account-settings/components/temp/Input.jsx
```diff
@@ -0,0 +1,105 @@
...
+    if (type === 'checkbox') return 'form-check-input';
+    if (type === 'radio') return 'form-check-input';
+    return 'form-control';
+  }
+
+  getRef(forwardedRef) {
+    if (process.env.NODE_ENV !== 'development') return forwardedRef;
```
albemarle: Do we do this frequently?  I wonder if it's worth having a util function for isProd().

src/account-settings/components/temp/Input.jsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+Input.defaultProps = {
+  className: undefined,
+  options: [],
+};
+
+
+// Using React.forwardRef – more on forwarding refs here:
+// https://reactjs.org/docs/forwarding-refs.html
```
albemarle: This comment would be more useful as "forwardRef floobs the wiginator in a crobley fashion".  I don't think the URL adds much; I am guessing Google will continue to exist at least as long as this code does.  ;-)

src/account-settings/reducers.js
```diff
@@ -0,0 +1,114 @@
...
+        loadingError: null,
+      };
+
+    case OPEN_FORM:
+      return {
+        ...state,
+        openFormId: action.payload.formId,
+      };
+    case CLOSE_FORM:
+      // Only close if the field to close is undefined or matches the field that is currently open
```
albemarle: Could this comment be eliminated with a variable name?

src/account-settings/AccountSettingsPage.jsx
abutterworth: Oh it looks like our pattern is the opposite (Sentence case for labels). Will update.

src/account-settings/components/temp/Input.jsx
davidjoy: This exists - it's in configuration as `ENVIRONMENT`.  The configuration is available in redux.  

In general we should use the configuration object and not process.env to keep a layer of abstraction.

src/account-settings/components/temp/Input.jsx
abutterworth: Good call. This component is a copy of one in a PR for Paragon. I'll update there.

src/account-settings/components/temp/Input.jsx
abutterworth: This component is a copy of one in a PR for Paragon. In Paragon, I think `process.env.NODE_ENV` is the safest thing to use. What do you think?

PR #6, opened 2019-04-25
Refactor and add email confirmation message

abutterworth:
Leaves the "field definition object" driven rendering behind for written out components for each field. It provides more flexibility. We can make an abstraction in the future if it's needed.

Saving an email now shows the proper confirmation message.
Screen Shot 2019-04-25 at 2 26 41 PM


src/account-settings/constants.js
```diff
@@ -0,0 +1,13 @@
+
+export const yearOfBirthOptions = (() => {
+  const currentYear = new Date().getFullYear();
+  const years = [];
+  let startYear = currentYear - 120;
```
douglashall: My goal in life is to one day make this code unusable for me.

abutterworth: Haha, good luck!

PR #9, opened 2019-04-26
Refactoring selectors a bit to organize/take advantage of reselect composition.

davidjoy:
Also fixing a few errors I saw w/r/t prop types.

### DIFF
--- a/src/account-settings/components/PasswordReset.jsx
+++ b/src/account-settings/components/PasswordReset.jsx
@@ -1,13 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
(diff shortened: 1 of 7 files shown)


src/account-settings/components/PasswordReset.jsx
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
```
abutterworth:
```suggestion
import { FormattedMessage } from 'react-intl';
import { injectIntl, intlShape } from '@edx/frontend-i18n';
```
Since we shim injectIntl to prevent errors we should import injectIntl and shape from @edx/frontend-i18n

davidjoy: Oh, right!! Thanks.  I auto-imported.

PR #10, opened 2019-04-29
feat: add linked accounts section

abutterworth:
Screen Shot 2019-04-29 at 4 43 00 PM

### DIFF
--- a/src/account-settings/components/ThirdPartyAuth.jsx
+++ b/src/account-settings/components/ThirdPartyAuth.jsx
@@ -20,7 +20,7 @@ class ThirdPartyAuth extends React.Component {
           <FormattedMessage
             id="account.settings.sso.link.account"
             defaultMessage="Sign in with {name}"
-            description="An action link to link a connected third party account"
+            description="An action link to link a connected third party account.m {name} will be Google, Facebook, etc."
             values={{ name }}
           />
         </Hyperlink>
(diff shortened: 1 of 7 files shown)


src/account-settings/components/ThirdPartyAuth.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+  renderConnectedProvider(url, name) {
+    return (
+      <React.Fragment>
+        <h6>{name}</h6>
+        <Hyperlink destination={url} className="btn btn-outline-primary">
+          <FormattedMessage
+            id="account.settings.sso.link.account"
+            defaultMessage="Sign in with {name}"
+            description="An action link to link a connected third party account"
```
albemarle: Can we add, "{name} will be Google, Facebook, etc." just in case the translators don't know what a connected third party account is?  (I think we should assume they're not necessarily technical folks.)

src/account-settings/components/ThirdPartyAuth.jsx
```diff
@@ -0,0 +1,139 @@
...
+      />
+    );
+  }
+
+  renderLoading() {
+    return (
+      <FormattedMessage
+        id="account.settings.sso.loading"
+        defaultMessage="Loading..."
+        description="Loading third party auth provider list"
```
albemarle: I think this can just be "we are waiting for some data to load" or something like that.

src/account-settings/components/ThirdPartyAuth.jsx
```diff
@@ -0,0 +1,139 @@
...
+        description="Error message for failing to load the third party auth provider list"
+      />
+    );
+  }
+
+  render() {
+    return (
+      <div>
+        <h2>Linked Accounts</h2>
+        <p>You can link your identity accounts to simplify signing in to edX.</p>
```
albemarle: i18n?

PR #11, opened 2019-04-30
feat: add social links

abutterworth:
Screen Shot 2019-04-30 at 2 34 51 PM

### DIFF
--- a/src/account-settings/service.js
+++ b/src/account-settings/service.js
@@ -7,7 +7,13 @@ let config = {
   PASSWORD_RESET_URL: null,
 };
 
-let apiClient = null; // eslint-disable-line no-unused-vars
+const SOCIAL_PLATFORMS = [
+  { id: 'twitter', key: 'social_link_twitter' },
+  { id: 'facebook', key: 'social_link_facebook' },
+  { id: 'linkedin', key: 'social_link_linkedin' },
+];
+
+let apiClient = null;
 
 function validateConfiguration(newConfig) {
   Object.keys(config).forEach((key) => {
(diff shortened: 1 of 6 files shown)


src/account-settings/service.js
```diff
@@ -17,30 +23,75 @@ function validateConfiguration(newConfig) {
...
+  const unpackedData = data;
+
+  SOCIAL_PLATFORMS.forEach(({ id, key }) => {
+    const platformData = data.social_links.find(({ platform }) => platform === id);
+    unpackedData[key] = typeof platformData === 'object' ? platformData.social_link : '';
+  });
+
+  if (Array.isArray(data.language_proficiencies)) {
+    if (data.language_proficiencies.length) {
+      unpackedData.language_proficiencies = data.language_proficiencies[0].code;
```
albemarle: We're only showing the first one?

abutterworth: Yes – we don't currently handle multi-select. If we want to change that we would have to update this code and the UI

src/account-settings/service.js
```diff
@@ -17,30 +23,75 @@ function validateConfiguration(newConfig) {
...
+    });
+  }
+  return Object.entries(unpackedFieldErrors)
+    .reduce((acc, [k, v]) => {
+      acc[k] = v.user_message;
+      return acc;
+    }, {});
+}
+
+function unpackAccountResponseData(data) {
```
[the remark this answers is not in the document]
abutterworth: It doesn't need to be – we'll address in the future as we discussed

PR #13, opened 2019-05-02
feat: improve confirmation messages on email and password

abutterworth:
Improve confirmation message handling
Screen Shot 2019-05-02 at 1 56 52 PM
- Makes sure shared strings do not use formattedmessage to avoid possibility of collisions


src/account-settings/components/Alert.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
davidjoy: Not the paragon component!?

src/account-settings/components/EditableField.jsx
```diff
@@ -1,14 +1,13 @@
...
 import { connect } from 'react-redux';
-import { FormattedMessage } from 'react-intl';
 import { injectIntl, intlShape } from '@edx/frontend-i18n'; // eslint-disable-line
 import { Button, StatefulButton } from '@edx/paragon';
 
 import Input from './temp/Input';
 import ValidationFormGroup from './temp/ValidationFormGroup';
 import SwitchContent from './temp/SwitchContent';
-import messages from './EditableField.messages';
+import messages from '../AccountSettingsPage.messages';
```
davidjoy: Hm, is it the right pattern to put everything into one messages file like this, rather than one per component where the message is used?  I feel like if we were going to put it all in one place, then maybe it should be named messages.jsx and kinda "belong" to the feature/module itself, you know?  I don't really feel strongly about splitting them up, except that putting it all in one place makes it marginally harder to extract sub-features/modules in the future.  (I'm used to having one messages file with thousands of entries, myself!)

src/account-settings/components/EditableField.jsx
```diff
@@ -123,11 +122,7 @@ function EditableField(props) {
                 onClick={handleCancel}
                 className="btn-outline-primary"
               >
-                <FormattedMessage
-                  id="account.settings.editable.field.action.cancel"
-                  defaultMessage="Cancel"
-                  description="The cancel button an editable field"
-                />
+                {intl.formatMessage(messages['account.settings.editable.field.action.cancel'])}
```
davidjoy: It's funny, I'm inclined to just do this all the time instead of using FormattedMessage... I'm not sure what value it really adds.

src/account-settings/components/PasswordReset.jsx
```diff
@@ -1,34 +1,44 @@
...
-              description="link text used in message: account.settings.editable.field.password.reset.button 'Contact technical support.'"
-            />
-          </Hyperlink>
-        ),
-      }}
-    />
+
+    <Alert
+      className="alert-warning mt-n2"
+      icon={<FontAwesomeIcon className="mr-2" icon={faExclamationTriangle} />}
```
davidjoy: Are these icons always going to be FontAwesomeIcons, or can we assume they are?  Could we pass in `faExclamationTriangle` here instead of JSX, maybe?

src/account-settings/components/EditableField.jsx
albemarle: Yeah, we were looking at that-- the issue is reusing strings in different components.  I don't mind this naming convention myself, assuming AccountSettingsPage is the lowest common ancestor of the files that need these strings.

src/account-settings/components/Alert.jsx
abutterworth: We don't have one yet

src/account-settings/components/EditableField.jsx
abutterworth: I've found in profile and here that there are occasionally strings that need to be shared among components. These actions, edit, cancel, save, etc. shouldn't really belong to any one component. I agree with a more generic name if we go the one messages file per module route.

src/account-settings/components/EditableField.jsx
abutterworth: Agreed, it also less likely we'll run into the colliding message id issue.

src/account-settings/components/PasswordReset.jsx
abutterworth: It's a good question – I think we could pull this Alert component into Paragon and use these icons as defaults? I wouldn't really want to assume that we'll receive a font awesome icon

PR #20, opened 2019-05-07
Site language dropdown - I hope.

davidjoy:
Where to begin?

1. It felt like the site language dropdown might become a reusable organism, so I attempted to model it as such.  In reality, the _component_ part of it isn't reusable at all because it's all tied up in how we do our forms on Account Settings.  Most everything else is reusable, however, so it's still in its own feature directory.

2. Making this work with EditableField required some refactoring of how that works - specifically to make it dumber and delegate to its parent on change and submit.  

3. The list of languages is hard coded in the client for v1 of i18n.

4. I deleted the local copy of Input.jsx.

5. The locale changes without a page refresh.

6. The options selectors for country and language proficiency moved into the selector so they could update as the locale changes.

7. I created some "utils" files in the common directory.  I anticipate these will be moved to frontend-utils in the future and will help us dry up our code a bit.

8. Because I had a hard coded list of languages, I took the liberty of updating i18n-loader to include all 14 languages in the dropdown... this is already out of date, as we're up to 21 languages as of this week.  Argh.

9. frontend-i18n (which is still in this repo) now exports some helper reducers and actions to keep track of and update the locale.  This enables number 5 above.

10. We do a gross hacky conversion of our client side two-letter locale codes back to the longer, more specific locales used by our server.  **If we ever have two versions of a language, this will immediately break.**

11. I updated handleRtl to be reversible in support of number 5.


package.json
```diff
@@ -31,7 +31,7 @@
     "@edx/frontend-component-footer": "^2.0.3",
     "@edx/frontend-component-site-header": "^2.1.4",
     "@edx/frontend-logging": "^2.0.0",
-    "@edx/paragon": "^4.1.3",
```
davidjoy: Necessary for the Input component.

src/account-settings/sagas.js
```diff
@@ -48,7 +50,18 @@ export function* handleSaveAccount(action) {
...
     const { commitValues, formId } = action.payload;
     const commitData = { [formId]: commitValues };
-    const savedValues = yield call(ApiService.patchAccount, username, commitData);
+    let savedValues = null;
+    if (formId === 'siteLanguage') {
+      yield all([
+        call(SiteLanguageApiService.patchPreferences, username, { prefLang: commitValues }),
+        call(SiteLanguageApiService.postSetLang, commitValues),
+      ]);
+      yield put(setLocale(commitValues));
```
abutterworth: Does the API provide a meaningful response to postSetLang that we should use instead?

src/account-settings/selectors.js
```diff
@@ -73,3 +73,71 @@ export const thirdPartyAuthSelector = createSelector(
     loadingError: accountSettings.thirdPartyAuthLoadingError,
   }),
 );
+
+/**
+ * If there's no draft present at all (undefined), use the original committed value.
```
abutterworth: Is this desired? The committed value was added to display a pending email change where the API will not update the email until an email link is clicked.

src/i18n/i18n-loader.js
```diff
@@ -20,40 +20,130 @@ import enLocale from 'react-intl/locale-data/en';
...
 import zhcnLocale from 'react-intl/locale-data/zh';
+import caLocale from 'react-intl/locale-data/ca';
+import heLocale from 'react-intl/locale-data/he';
+import idLocale from 'react-intl/locale-data/id';
+import kokrLocale from 'react-intl/locale-data/ko';
+import plLocale from 'react-intl/locale-data/pl';
+import ptbrLocale from 'react-intl/locale-data/pt';
+import ruLocale from 'react-intl/locale-data/ru';
+import thLocale from 'react-intl/locale-data/th';
+import ukLocale from 'react-intl/locale-data/uk';
```
abutterworth: Wow!

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -47,9 +39,23 @@ class AccountSettingsPage extends React.Component {
   componentDidMount() {
     this.props.fetchAccount();
     this.props.fetchThirdPartyAuthProviders();
+    this.props.fetchSiteLanguages();
```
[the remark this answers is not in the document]
davidjoy: Yeah, truth.  This could be bundled into a single action call here that fans out to multiple actions invoked by a saga, for instance.  Alternately, as per some of our performance conversations, we might want to find a way to just do these calls before React does all its component mounting, or at least in parallel with it.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -60,69 +66,100 @@ class AccountSettingsPage extends React.Component {
...
+              options={this.props.siteLanguageOptions}
+              value={this.props.siteLanguage}
+              label={this.props.intl.formatMessage(messages['account.settings.field.site.language'])}
+              {...editableFieldProps}
+            />
             <EditableField
               name="country"
               type="select"
-              options={this.countryOptions}
+              value={this.props.formValues.country}
```
albemarle: For the future, it would be even more awesome if separate stuff had its own PR, especially since this one is already long.  And then you could have better commit messages (I assume this one would have been a fix: .... ?).

src/account-settings/sagas.js
davidjoy: Unfortunately not.  It actually performs a redirect that we're ignoring here - it works, but it wasn't quite built to be called from AJAX.

src/account-settings/selectors.js
```diff
@@ -73,3 +73,71 @@ export const thirdPartyAuthSelector = createSelector(
...
+const languageProficiencyOptionsSelector = createSelector(
+  localeSelector,
+  locale => getLanguageList(locale).map(({ code, name }) => ({ value: code, label: name })),
+);
+
+
+/**
+ * This selector converts the site language code back to the server version so that it can match up
+ * with one of the options in the site language dropdown.  The drafts version will already be the
+ * server version, but if it's from localeSelector, it will be our client (two character) version.
```
albemarle: Good comment.

src/account-settings/selectors.js
davidjoy: This is what chooses to include the draft value if it exists or the original value if it doesn't - I grabbed this code from frontend-app-profile to build formValuesSelector below.  The word committed is unrelated to the email change...

src/i18n/i18n-loader.js
```diff
@@ -79,19 +169,25 @@ const rtlLocales = ['ar', 'he', 'fa', 'ur'];
 const isRtl = locale => rtlLocales.includes(locale);
 
 const handleRtl = () => {
-  if (isRtl(getLocale())) {
-    document.getElementsByTagName('html')[0].setAttribute('dir', 'rtl');
-    document.styleSheets[0].disabled = true;
-  } else {
-    document.styleSheets[1].disabled = true;
+  if (process.env.ENVIRONMENT === 'production') {
```
albemarle: Did we ever add an isProduction() function?

This could be a good place for a comment as to why we're (temporarily) not doing this in dev.

src/i18n/i18n-loader.js
```diff
@@ -79,19 +169,25 @@ const rtlLocales = ['ar', 'he', 'fa', 'ur'];
...
   }
 };
 
 /**
  * Provides a lookup table of country IDs to country names for the current locale.
  */
 const getCountryMessages = (locale) => {
-  const finalLocale = countryLangs().includes(locale) ? locale : 'en';
+  const twoLetterLocale = getTwoLetterLanguageCode(locale);
+  const finalLocale = countryLangs().includes(twoLetterLocale) ? twoLetterLocale : 'en';
```
albemarle: serverLocale?

src/account-settings/AccountSettingsPage.jsx
davidjoy: This is a change related to the PR, not a fix.  

`this.countryOptions` was defined in the constructor of the react component, which meant that it doesn't update if the locale changes.  By moving it to props/having it defined in a selector, it can respond to the locale change and update the country list to show the names in the newly-selected language.

albemarle:
Would you mind running ``make i18n.extract && make i18n.concat`` and also committing the updated ``transifex_input.json`` file?

src/account-settings/AccountSettingsPage.jsx
davidjoy: If something in the review isn't clear, I'm happy to talk it through.  I don't expect reviews to always be easy to throw over a wall, and honestly prefer to look at them in person.

src/i18n/i18n-loader.js
davidjoy: I actually... cheated... here a bit.  We have this environment variable in our configuration, which gets put into redux/made available through it to whatever components want to see it.  If you're not a component, though, we've been using initialization/configuration time (index.jsx at the top of the repo) as a place to pass data to our various supporting libraries (auth, logging, analytics, etc.) Up until now, i18n didn't need that and so had no need to be configured.  I didn't want to bite that off in this PR, so I just snagged it out of process.env like a horrible person.  Ideally i18n would have a "configure" method which would get called with whatever information it needs to run itself, including this flag, thus isolating it from knowledge of where that configuration came from or who's responsible for it.

davidjoy: Honestly, I think I'd like to change that before finishing off the i18n repo card.

src/i18n/i18n-loader.js
davidjoy: In this case it's not so much about server as about choosing the locale we actually want to do the lookup with, which may need to fall back to 'en'.  (And, actually, since it's the two letter one, it's explicitly _not_ the server locale here, come to think of it. 😄)

PR #22, opened 2019-05-08
feat: add time zone

abutterworth:
- Adds the time zone field which edits the User Preference (not Account).
- Consolidates fetching of data under the name FETCH_SETTINGS, fetchSettings, etc.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -8,7 +8,7 @@ import {
 
 import messages from './AccountSettingsPage.messages';
 
-import { fetchAccount, fetchThirdPartyAuthProviders, updateDraft, saveAccount } from './actions';
+import { fetchSettings, saveSettings, updateDraft } from './actions';
 import { accountSettingsPageSelector } from './selectors';
 
 import { PageLoading } from '../common';
@@ -38,13 +38,12 @@ class AccountSettingsPage extends React.Component {
     this.timeZoneOptions = Array.concat(
       [{ value: '', label: Intl.DateTimeFormat().resolvedOptions().timeZone }],
       // eslint-disable-next-line no-unused-vars
-      TIME_ZONES.map(([label, value]) => ({ value, label: value })),
+      TIME_ZONES.map(([value, label]) => ({ value, label })),
     );
   }
 
   componentDidMount() {
-    this.props.fetchAccount();
-    this.props.fetchThirdPartyAuthProviders();
+    this.props.fetchSettings();
     this.props.fetchSiteLanguages();
   }
 
@@ -53,7 +52,7 @@ class AccountSettingsPage extends React.Component {
   }
 
   handleSubmit = (formId, values) => {
-    this.props.saveAccount(formId, values);
+    this.props.saveSettings(formId, values);
   }
 
   renderContent() {
@@ -174,7 +173,6 @@ class AccountSettingsPage extends React.Component {
               label={this.props.intl.formatMessage(messages['account.settings.field.time.zone'])}
               {...editableFieldProps}
             />
-
           </div>
         </div>
       </div>
@@ -252,11 +250,10 @@ AccountSettingsPage.propTypes = {
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   })),
 
-  fetchAccount: PropTypes.func.isRequired,
-  fetchThirdPartyAuthProviders: PropTypes.func.isRequired,
   fetchSiteLanguages: PropTypes.func.isRequired,
   updateDraft: PropTypes.func.isRequired,
-  saveAccount: PropTypes.func.isRequired,
+  saveSettings: PropTypes.func.isRequired,
+  fetchSettings: PropTypes.func.isRequired,
 };
 
 AccountSettingsPage.defaultProps = {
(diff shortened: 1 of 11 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -34,11 +35,15 @@ class AccountSettingsPage extends React.Component {
       value: key,
       label: props.intl.formatMessage(messages[`account.settings.field.gender.options.${key}`]),
     }));
+    this.timeZoneOptions = Array.concat(
```
davidjoy: Does this change when the locale changes?  If so, you can move it to the selector so that it gets re-evaluated when the user updates their site language.  That's what I had to do with countryOptions and languageProficiencyOptions to get them to work.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -160,7 +165,14 @@ class AccountSettingsPage extends React.Component {
               label={this.props.intl.formatMessage(messages['account.settings.field.social.platform.name.twitter'])}
               {...editableFieldProps}
             />
-
+            <EditableField
+              name="time_zone"
+              type="select"
+              value={this.props.formValues.time_zone || ''}
+              options={this.timeZoneOptions}
+              label={this.props.intl.formatMessage(messages['account.settings.field.time.zone'])}
```
davidjoy: The timezone form element on the old page has help text: "Select the time zone for displaying course dates. If you do not specify a time zone, course dates, including assignment deadlines, will be displayed in your browser's local time zone."

src/account-settings/reducers.js
```diff
@@ -134,35 +134,35 @@ const accountSettingsReducer = (state = defaultState, action) => {
...
+    //     thirdPartyAuthLoaded: false,
+    //     thirdPartyAuthLoadingError: action.payload.error,
+    //   };
+    // case FETCH_THIRD_PARTY_AUTH_PROVIDERS.RESET:
+    //   return {
+    //     ...state,
+    //     thirdPartyAuthLoading: false,
+    //     thirdPartyAuthLoaded: false,
+    //     thirdPartyAuthLoadingError: null,
+    //   };
```
albemarle: Left in on purpose?

src/account-settings/components/ThirdPartyAuth.jsx
```diff
@@ -4,14 +4,9 @@ import { connect } from 'react-redux';
 import { FormattedMessage } from 'react-intl';
 import { Hyperlink } from '@edx/paragon';
 
-import { fetchThirdPartyAuthProviders } from '../actions';
 import { thirdPartyAuthSelector } from '../selectors';
 
 class ThirdPartyAuth extends React.Component {
-  componentDidMount() {
-    this.props.fetchThirdPartyAuthProviders();
```
davidjoy: Yikes, were we basically doing this twice?

src/account-settings/service.js
```diff
@@ -116,3 +127,53 @@ export async function getThirdPartyAuthProviders() {
...
+  if (!isEmpty(accountCommitValues)) {
+    patchRequests.push(patchAccount(username, accountCommitValues));
+  }
+  if (!isEmpty(preferenceCommitValues)) {
+    patchRequests.push(patchPreferences(username, preferenceCommitValues));
+  }
+
+  const results = await Promise.all(patchRequests);
+  // Assigns in order of requests. Preference keys
+  // will override account keys. Notably time_zone.
```
albemarle: Do you mean there is both an account setting & a user preference for timezone?

src/account-settings/constants/TIME_ZONES.js
```diff
@@ -0,0 +1,1764 @@
+// This list of timezones is copied from the
+// account settings page on edx platform.
+export const TIME_ZONES = [ // eslint-disable-line import/prefer-default-export
```
davidjoy: Looking at the existing account settings, it has the timezones broken up into two groups - "Country time zones" and "All time zones".  Not sure how important that is, but we might want to clear it with Marco if we're losing it...

src/account-settings/reducers.js
```diff
@@ -134,35 +134,35 @@ const accountSettingsReducer = (state = defaultState, action) => {
...
-        thirdPartyAuthLoadingError: action.payload.error,
-      };
-    case FETCH_THIRD_PARTY_AUTH_PROVIDERS.RESET:
-      return {
-        ...state,
-        thirdPartyAuthLoading: false,
-        thirdPartyAuthLoaded: false,
-        thirdPartyAuthLoadingError: null,
-      };
+    // case FETCH_THIRD_PARTY_AUTH_PROVIDERS.BEGIN:
```
davidjoy: Delete me!

src/account-settings/reducers.js
abutterworth: Nope

src/account-settings/AccountSettingsPage.jsx
abutterworth: It doesn't yet have any localized versions, should in the future I hope. For now though it seems more important to make sure this list matches edx-platform.

src/account-settings/components/ThirdPartyAuth.jsx
abutterworth: Looks like we were :P

src/account-settings/constants/TIME_ZONES.js
abutterworth: Good call will check.

src/account-settings/reducers.js
abutterworth: 🔨

src/account-settings/AccountSettingsPage.jsx
abutterworth: Added

src/account-settings/service.js
abutterworth: Added comment slightly above regarding our conversation.

PR #23, opened 2019-05-09
Make certain field uneditable for enterprise learners.

douglashall:


src/account-settings/selectors.js
```diff
@@ -74,6 +76,16 @@ export const thirdPartyAuthSelector = createSelector(
...
 );
 
+export const profileDataManagerSelector = createSelector(
+  accountSettingsSelector,
+  accountSettings => accountSettings.profileDataManager,
+);
+
+export const staticFieldsSelector = createSelector(
+  accountSettingsSelector,
+  accountSettings => (accountSettings.profileDataManager ? ['email', 'country'] : []),
```
robrap: The old code also protected Full Name. Should that be added here?

PR #25, opened 2019-05-10
fix: verify parity

abutterworth:
- Change order of fields, update edx-bootstrap to latest.
- Password reset now sends email along with post request.
- Clicks on "reset password" now a noop while in progress.
- add best guesses for help text

Screen Shot 2019-05-10 at 12 29 02 PM
Screen Shot 2019-05-10 at 12 31 08 PM

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -68,7 +68,9 @@ class AccountSettingsPage extends React.Component {
       <div>
         <div className="row">
           <div className="col-md-8 col-lg-6">
-            <h2>{this.props.intl.formatMessage(messages['account.settings.section.account.information'])}</h2>
+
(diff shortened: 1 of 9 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -68,14 +68,17 @@ class AccountSettingsPage extends React.Component {
       <div>
         <div className="row">
           <div className="col-md-8 col-lg-6">
-            <h2>{this.props.intl.formatMessage(messages['account.settings.section.account.information'])}</h2>
+
+
```
douglashall: Were these extra lines intentional?

abutterworth: Yea trying to make it clearer what is a headline/grouping

PR #26, opened 2019-05-10
ARCH-785: add analytics event edx.user.settings.viewed

robrap:
ARCH-785

I'm leaning towards skipping a test (there are currently no tests in this file).  The legacy app didn't have one for this event, and not sure why researchers would ever care about this event anyway. :)

Given that, if someone thinks it is important, just say so.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,7 +1,13 @@
 import 'babel-polyfill';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { identifyAuthenticatedUser, sendPageEvent, configureAnalytics, initializeSegment } from '@edx/frontend-analytics';
+import {
+  configureAnalytics,
+  identifyAuthenticatedUser,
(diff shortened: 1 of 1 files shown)


src/index.jsx
```diff
@@ -62,11 +68,18 @@ apiClient.ensurePublicOrAuthenticationAndCookies(
...
 
     ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
 
     identifyAuthenticatedUser();
     sendPageEvent();
+
+    sendTrackingLogEvent('edx.user.settings.viewed', {
+      page: 'account',
+      visibility: null,
+      user_id: authState.authentication && authState.authentication.userId,
```
albemarle: Just to check, you're intending this to be true or false?  Or are you trying to send the userId if it exists?

robrap: User id. I will test before shipping.

PR #28, opened 2019-05-10
Add recovery email field.

douglashall:


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -92,6 +92,20 @@ class AccountSettingsPage extends React.Component {
     );
   }
 
+  renderSecondaryEmailField() {
+    if (this.props.hiddenFields.includes('secondary_email')) {
+      return (null);
```
abutterworth: No parens?
```suggestion
      return null;
```

src/account-settings/selectors.js
```diff
@@ -86,6 +86,11 @@ export const staticFieldsSelector = createSelector(
   accountSettings => (accountSettings.profileDataManager ? ['name', 'email', 'country'] : []),
 );
 
+export const hiddenFieldsSelector = createSelector(
+  accountSettingsSelector,
+  accountSettings => (accountSettings.profileDataManager ? [] : ['secondary_email']),
```
abutterworth: Am I reading this right? If there's no profileDataManager then there are no hidden fields?

douglashall: Correct. The presence of a profileDataManager indicates that we are dealing with an enterprise user, so we need to show the `secondary_email` field.

PR #29, opened 2019-05-10
fix: saving an account setting will no longer delete a time zone pref…

abutterworth:
The account endpoint returns a null, uneditable time_zone value which collides with our preferences value when saving an account setting. This deletes time_zone when unpacking/transforming account response data.

### DIFF
--- a/src/account-settings/service.js
+++ b/src/account-settings/service.js
@@ -41,6 +41,9 @@ function unpackFieldErrors(fieldErrors) {
 function unpackAccountResponseData(data) {
   const unpackedData = data;
 
+  // This is handled by preferences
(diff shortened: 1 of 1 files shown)


src/account-settings/service.js
```diff
@@ -41,6 +41,9 @@ function unpackFieldErrors(fieldErrors) {
 function unpackAccountResponseData(data) {
   const unpackedData = data;
 
+  // This is handled by preferences
+  if (unpackedData.time_zone !== undefined) delete unpackedData.time_zone;
```
albemarle: I think you don't even need the if here.

PR #36, opened 2019-05-13
fix: send social link value when empty string is supplied

abutterworth:
Makes it possible to delete a social link.


src/account-settings/service.js
```diff
@@ -63,7 +63,7 @@ function packAccountCommitData(commitData) {
   const packedData = commitData;
 
   SOCIAL_PLATFORMS.forEach(({ id, key }) => {
-    if (commitData[key]) {
+    if (commitData[key] !== undefined) {
```
albemarle: Could you add a brief comment that this is so an empty string can delete it?

PR #37, opened 2019-05-13
feat: delete my account

albemarle:
Still needs:

- Add "you must enter a password" error state on confirmation modal if the password field loses focus while it's blank-- I propose doing this as a fast follow.

Note: Seeing console error: "Each child in a list should have a unique "key" prop."  (@abutterworth this looks like a Paragon thing with Modal's Buttons)

### DIFF
--- a/src/account-settings/components/DeleteMyAccount.jsx
+++ b/src/account-settings/components/DeleteMyAccount.jsx
@@ -1,0 +1,248 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { injectIntl, intlShape } from '@edx/frontend-i18n'; // eslint-disable-line
+import { Button, Hyperlink, Input, Modal, ValidationFormGroup } from '@edx/paragon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faExclamationCircle, faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import { FormattedMessage } from 'react-intl';
+import { configuration } from '../../environment';
+import Alert from './Alert';
+import { deleteAccount, deleteAccountConfirmation, deleteAccountReset } from '../actions';
+import messages from '../AccountSettingsPage.messages';
+
+const passwordFieldId = 'passwordFieldId';
+
+class DeleteMyAccount extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      password: '',
+    };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleCancel = this.handleCancel.bind(this);
+    this.handlePasswordChange = this.handlePasswordChange.bind(this);
+    this.handleFinalClose = this.handleFinalClose.bind(this);
+  }
+
+  handleSubmit() {
+    this.props.deleteAccount(this.state.password);
+  }
+
+  handleCancel() {
+    this.setState({ password: '' });
+    this.props.deleteAccountReset();
+  }
+
+  handlePasswordChange(e) {
+    this.setState({ password: e.target.value });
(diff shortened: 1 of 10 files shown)


abutterworth:
The console error you're seeing shouldn't affect the app at all, I'm working on refactoring the modal now and this will go away when that's done. Maybe we can live with the console error for some time?

src/account-settings/reducers.js
```diff
@@ -137,6 +138,13 @@ const accountSettingsReducer = (state = defaultState, action) => {
         ...state,
         previousSiteLanguage: action.payload.previousSiteLanguage,
       };
+
+    case DELETE_ACCOUNT.FAILURE:
+      return {
+        ...state,
+        passwordError: true,
```
abutterworth: I would consider naming this `deleteAccountError` or something instead.

src/account-settings/components/DeleteMyAccount.jsx
```diff
@@ -0,0 +1,248 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { injectIntl, intlShape } from '@edx/frontend-i18n'; // eslint-disable-line
+import { Button, Hyperlink, Input, Modal, ValidationFormGroup } from '@edx/paragon';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faExclamationCircle, faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
+import { FormattedMessage } from 'react-intl';
+import { configuration } from '../../environment';
```
davidjoy: So for this - The redux store has the "configuration" object in it at `state.configuration` - therefore the selector in AccountSettingsPage.jsx could just add in the variables it and this child need and pass it down as a prop. 

I noticed as part of looking this up that AccountSettingsPage.jsx is currently importing the environment configuration itself, rather than finding it in the store, which isn't your code, but fixing it for this would also fix it there, fwiw.

src/account-settings/sagas.js
```diff
@@ -1,7 +1,7 @@
 
 import { call, put, delay, takeEvery, select, all } from 'redux-saga/effects';
 import { push } from 'connected-react-router';
-import { logAPIErrorResponse } from '@edx/frontend-logging';
```
davidjoy: I think this change will blow up - LoggingService isn't exported anymore, I don't think... calling that second import there, logAPIErrorResponse, is the correct way to do it. 

(The above is because frontend-logging now gets configured with a NewRelicLoggingService (which used to be called LoggingService) and then exposes the individual methods as its interface, rather than exposing what used to be called LoggingService directly)

src/account-settings/components/DeleteMyAccount.jsx
albemarle: @davidjoy Are you okay with that change going in the fast follow?

src/account-settings/sagas.js
```diff
@@ -98,6 +102,21 @@ export function* handleSaveSettings(action) {
...
+export function* handleDeleteAccount(action) {
+  try {
+    yield put(deleteAccountBegin());
+    const response = yield call(ApiService.postDeleteAccount, action.payload.password);
+    yield put(deleteAccountSuccess(response));
+  } catch (e) {
+    if (typeof e.response.data === 'string') {
+      yield put(deleteAccountFailure());
+    } else {
+      LoggingService.logAPIErrorResponse(e);
```
davidjoy: So as per the above comment - this line will blow up with a "cannot call logAPIErrorResponse of undefined"

src/account-settings/sagas.js
albemarle: D'oh, thanks for the catch!

PR #41, opened 2019-05-14
Display duplicate TPA provider message.

douglashall:


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -73,6 +73,27 @@ class AccountSettingsPage extends React.Component {
...
+      return null;
+    }
+
+    return (
+      <div>
+        <Alert className="alert alert-danger" role="alert">
+          <FormattedMessage
+            id="account.settings.message.duplicate.tpa.provider"
+            defaultMessage="The {provider} account you selected is already linked to another edX account."
+            description="alert message informing the user that the third-party account they attempted to link is already linked to another edX account"
```
nasthagiri: Newbie question: Why are some strings defined inline (like here) while some are defined separately in [AccountSettingsPage.messages.jsx](https://github.com/edx/frontend-app-account/blob/master/src/account-settings/AccountSettingsPage.messages.jsx)?

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -134,7 +155,7 @@ class AccountSettingsPage extends React.Component {
       <div>
         <div className="row">
           <div className="col-md-8 col-lg-6">
-
+            {this.renderDuplicateTpaProviderMessage()}
```
nasthagiri: Should this error message be co-located with the section it corresponds to?  Namely, within the `Linked Accounts` section?

Or does this error occur when the page is loaded - and not when a `Linked Account` is activated?

src/account-settings/selectors.js
```diff
@@ -95,6 +95,21 @@ export const hiddenFieldsSelector = createSelector(
   accountSettings => (accountSettings.profileDataManager ? [] : ['secondary_email']),
 );
 
+export const duplicateTpaProviderSelector = createSelector(
+  accountSettingsSelector,
+  () => {
+    const url = new URL(global.location.href);
```
nasthagiri: Newbie question: Is the use of `global` gatsby-compatible?

src/account-settings/AccountSettingsPage.jsx
douglashall: I think it depends on whether you are using the `FormattedMessage` component vs. that helper function we use when assigning strings to attributes. @albemarle might be a better person to ask.

src/account-settings/AccountSettingsPage.jsx
douglashall: It is displayed when the page is loaded. The user would not see it if it was near the Linked Account section because it would be below the fold.

src/account-settings/selectors.js
douglashall: Yes, this is the thing that makes it gatsby-compatible.

douglashall: On second thought, I have replaced the use of `global` here with `window` since this is not and will not be a gatsby app. Using `global` really only applies to code that could potentially be included in a gatsby build.

nasthagiri: I believe the account settings page will be used in the Enterprise learner portal.  But @georgebabey can confirm.

georgebabey: Yes, the plan is to eventually pull in the page components for these pages into the branded build. I'm told for masters rollout, because the header is universal, this isn't needed but once we work on the learner portal for enterprise it will be.

Was there complications in leaving it as global?

douglashall: Yes, it wasn't working for me. Probably a simple fix with the way I was configuring webpack, but I didn't want to block this work on figuring it out right now. Happy to talk about it if/when this becomes a blocker.

PR #54, opened 2019-05-20
Use frontend-i18n package instead of local i18n directory.

davidjoy:
Delete the local copy of the i18n library.  

NOTE: the package.json path MUST be updated prior to merging this PR.


package.json
```diff
@@ -30,6 +30,7 @@
     "@edx/frontend-auth": "^5.3.0",
     "@edx/frontend-component-footer": "^4.1.4",
     "@edx/frontend-component-site-header": "^2.1.6",
+    "@edx/frontend-i18n": "file:../frontend-i18n",
```
davidjoy: This must be updated prior to merging.  Waiting for frontend-i18n to be published to npm, which is waiting on a DevOps ticket.

src/index.jsx
```diff
@@ -53,6 +54,8 @@ function createInitialState() {
 }
 
 function configure() {
+  configureI18n(configuration, messages);
```
davidjoy: This is the new way of setting up i18n.  You configure the library with config that it needs and your app's message JSON files.

src/site-language/service.js
```diff
@@ -36,7 +34,7 @@ async function patchPreferences(username, params) {
 
 async function postSetLang(code) {
   const formData = new FormData();
-  formData.append('language', getAssumedServerLanguageCode(code));
+  formData.append('language', code);
```
davidjoy: Because the client and server now agree on language codes, the function that was here is no longer necessary.

src/i18n/index.js
```diff
@@ -0,0 +1,32 @@
...
+  'zh-cn': zhcnMessages,
+  ca: caMessages,
+  he: heMessages,
+  id: idMessages,
+  'ko-kr': kokrMessages,
+  pl: plMessages,
+  'pt-br': ptbrMessages,
+  ru: ruMessages,
+  th: thMessages,
+  uk: ukMessages,
```
albemarle: I have a mild preference for quoting all keys if some of them need quoting, but only a mild one.

davidjoy: The linter complained.

PR #67, opened 2019-05-28
Add empty states and ability to delete values from select fields

abutterworth:
Add a clickable placeholder for empty fields. For empty static fields, show a message indicating no value is set.

All select fields have placeholder options that can be selected to delete the value.

Screen Shot 2019-05-28 at 2 28 13 PM

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -36,12 +36,24 @@ class AccountSettingsPage extends React.Component {
     super(props);
     this.educationLevels = EDUCATION_LEVELS.map(key => ({
       value: key,
-      label: props.intl.formatMessage(messages[`account.settings.field.education.levels.${key}`]),
+      label: props.intl.formatMessage(messages[`account.settings.field.education.levels.${key || 'empty'}`]),
     }));
     this.genderOptions = GENDER_OPTIONS.map(key => ({
       value: key,
-      label: props.intl.formatMessage(messages[`account.settings.field.gender.options.${key}`]),
+      label: props.intl.formatMessage(messages[`account.settings.field.gender.options.${key || 'empty'}`]),
     }));
+    this.languageProficiencyOptions = [{
+      value: '',
+      label: props.intl.formatMessage(messages['account.settings.field.language_proficiencies.options.empty']),
+    }].concat(props.languageProficiencyOptions);
+    this.yearOfBirthOptions = [{
+      value: '',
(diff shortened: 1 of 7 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -67,6 +79,10 @@ class AccountSettingsPage extends React.Component {
     return concatTimeZoneOptions;
   });
 
+  getIsEditable(fieldName) {
```
albemarle: 👍 

Could also be called just isEditable.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -126,6 +142,15 @@ class AccountSettingsPage extends React.Component {
     );
   }
 
+  renderEmptyStaticFieldMessage() {
+    if (!this.props.profileDataManager) {
```
albemarle: Do you think it's worth making a isEnterprise function?  Without knowing what this is, I don't think I'd know what this is.  ;-)

src/account-settings/components/EditableField.jsx
```diff
@@ -66,6 +57,26 @@ function EditableField(props) {
     onCancel(name);
   };
 
+  const renderEmptyLabel = () => {
+    if (!isEditable) {
```
albemarle: If you have both an if and an else, it can be nice to make the if condition not unpositive.  ;-)

src/account-settings/components/EditableField.jsx
```diff
@@ -66,6 +57,26 @@ function EditableField(props) {
...
+    }
+    return <Button onClick={handleEdit} className="btn-link p-0">{emptyLabel}</Button>;
+  };
+
+  const renderValue = (rawValue) => {
+    if (!rawValue) return renderEmptyLabel();
+
+    if (options) {
+      // Use == instead of === to prevent issues when HTML casts numbers as strings
+      // eslint-disable-next-line eqeqeq
```
albemarle: eqeqeq makes me look around for a 🐁

albemarle:
erase

src/account-settings/AccountSettingsPage.jsx
abutterworth: Made an update containing a `isManagedProfile` method.

src/account-settings/components/EditableField.jsx
abutterworth: I'll try to stay positive :p

PR #70, opened 2019-05-29
feat(delete my account): improvements in error handling for confirmation modal

albemarle:
- Button is now always clickable (addressing ARCH-839)
- Submitting with a blank password now shows an error
- When an error is showing, typing in the input clears the error

### DIFF
--- a/src/account-settings/components/DeleteMyAccount.jsx
+++ b/src/account-settings/components/DeleteMyAccount.jsx
@@ -5,7 +5,6 @@ import { FormattedMessage, injectIntl, intlShape } from '@edx/frontend-i18n';
 import { Button, Hyperlink, Input, Modal, ValidationFormGroup } from '@edx/paragon';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faExclamationCircle, faExclamationTriangle } from '@fortawesome/free-solid-svg-icons';
-import trim from 'lodash.trim';
 
 import Alert from './Alert';
 import { deleteAccount, deleteAccountConfirmation, deleteAccountFailure, deleteAccountReset, deleteAccountCancel } from '../actions';
@@ -51,7 +50,7 @@ class DeleteMyAccount extends React.Component {
   }
 
   handlePasswordChange(e) {
-    this.setState({ password: trim(e.target.value) });
+    this.setState({ password: e.target.value.trim() });
     this.props.deleteAccountReset();
   }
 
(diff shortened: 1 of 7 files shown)


src/account-settings/components/DeleteMyAccount.jsx
```diff
@@ -24,17 +25,34 @@ class DeleteMyAccount extends React.Component {
     this.handleFinalClose = this.handleFinalClose.bind(this);
   }
 
+  /**
+   * @returns String The message id for a short description of the error, suitable for a header or as the error message under an input field.
+   */
+  getShortErrorMessageId(reason) {
```
abutterworth: Will passing/fetching message id's around like this become problematic in anyway down the road?

src/account-settings/components/DeleteMyAccount.jsx
```diff
@@ -24,17 +25,34 @@ class DeleteMyAccount extends React.Component {
...
 
   handleCancel() {
     this.setState({ password: '' });
-    this.props.deleteAccountReset();
+    this.props.deleteAccountCancel();
   }
 
   handlePasswordChange(e) {
-    this.setState({ password: e.target.value });
+    this.setState({ password: trim(e.target.value) });
```
abutterworth: would `e.target.value.trim()` work?

src/account-settings/reducers.js
```diff
@@ -162,10 +162,17 @@ const accountSettingsReducer = (state = defaultState, action) => {
       return {
         ...state,
         accountDeletionState: 'failed',
-        deletionError: 'server',
+        deletionError: action.payload.reason || 'server',
```
abutterworth: Just a note that I think this is different from other places in the application where error is usually a string of the error, not its type.

src/account-settings/components/DeleteMyAccount.jsx
albemarle: I went back and forth on passing around the id or the string itself, but since in some methods in this file it was clearly preferable to pass the id and not the translated string, I decided to stick with that for all of them.

src/account-settings/components/DeleteMyAccount.jsx
albemarle: According to this [yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim).  I'm so out of the habit of using the String methods because they are (were?) more likely than average to cause problems on IE.

src/account-settings/reducers.js
albemarle: Would deletionErrorType be clearer?

PR #75, opened 2019-05-30
fix(i18n): tell babel-plugin-react-intl where defineMessages and friends are now being imported from

albemarle:


robrap:
@albemarle: Nice find! Can we add the safety checks to the Makefile before turning on jobs for other repos?  It can be a separate PR.  Let me know if you want to discuss. Thanks.

albemarle:
@robrap Yup, I'll have separate PRs for some Makefile improvements.

PR #81, opened 2019-06-03
Tidy: step 1, shuffle common components and add a utility function

davidjoy:
Taking this one step at a time.  If this loads and we see the following components render, everything's fine:

- BetaLanguageBanner
- DeleteMyAccount
- JumpNav
- PasswordReset
- EmailField (two of these)
- ThirdPartyAuth
- EditableField (most of the "Edit"-able fields)
- SwitchContent (open any edit button)
- Alert (used by AccountSettingsPage, BetaLanguageBanner, DeleteMyAccount, EmailField, PasswordReset, and ThirdPartyAuth) 

This PR also includes a utility function I intend to use as I modularize the above.


src/common/__snapshots__/utils.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
albemarle: Just checking that this isn't leftover?

src/common/utils.js
```diff
@@ -47,6 +47,37 @@ export function keepKeys(data, whitelist) {
...
+ */
+export function getModuleState(state, path) {
+  if (path.length < 1) {
+    return state;
+  }
+  const key = path.shift();
+  if (state[key] === undefined) {
+    throw new Error(`Unexpected state key ${key} given to getModuleState. Is your state path set up correctly?`);
+  }
+  return getModuleState(state[key], path);
```
albemarle: Are you sure this will never have cycles?  Maybe it would be good to add a maxDepth counter.

davidjoy: It can't really have cycles... It takes an array and shifts off the first element on each iteration, and its end condition is that the array length is less than one.  Otherwise, it has no side effects that might add elements to the array (nor should any code ever be modifying that array to begin with)

Even if there were a cycle in state - which there shouldn't be or someone has made a massive mistake - it's only going to traverse it until it runs out of path elements.

PR #92, opened 2019-06-05
Modularizing ResetPassword.

davidjoy:
# Testing

- [ ] Click the reset password button.  It should show a little spinner while it works.  It should show you the reset password confirmation alert and email you.


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -19,7 +19,7 @@ import { Alert, PageLoading } from '../common';
 import JumpNav from './JumpNav';
 import DeleteAccount from './delete-account';
 import EditableField from './EditableField';
-import PasswordReset from './PasswordReset';
+import ResetPassword from './reset-password';
```
albemarle: 👍

src/account-settings/reset-password/data/actions.js
```diff
@@ -0,0 +1,24 @@
+import { utils } from '../../../common';
```
albemarle: Ouch.

src/account-settings/reset-password/data/actions.js
```diff
@@ -0,0 +1,24 @@
...
+
+const { AsyncActionType } = utils;
+
+export const RESET_PASSWORD = new AsyncActionType('ACCOUNT_SETTINGS', 'RESET_PASSWORD');
+
+// RESET PASSWORD ACTIONS
```
albemarle: Can delete all these organizing comments now!

src/account-settings/reset-password/data/actions.js
davidjoy: Yeah, I hear you.  In reality, everything in that directory is stuff I'm envisioning being put into `frontend-utils` and then shared across MFEs, so it's not as crazy as it seems. :D

src/account-settings/reset-password/data/actions.js
davidjoy: Whoops!  Yes.  Was trying to do that. :) . Will remove.

src/account-settings/reducers.js
```diff
@@ -223,6 +215,13 @@ const reducer = (state = defaultState, action) => {
         siteLanguage: siteLanguageReducer(state.siteLanguage, action),
       };
 
+    case RESET_PASSWORD.BEGIN:
+    case RESET_PASSWORD.SUCCESS:
```
abutterworth: `combineReducers` it that makes sense?

davidjoy: Yeah - trying to get to the point with this where it can use combine reducers, but was trying to avoid refactoring the whole thing at once.  (I put a comment to that effect in one of these PRs...)

PR #93, opened 2019-06-05
Modularizing third party auth.

davidjoy:
NOTE: At present, I couldn't figure out how to test this in devstack.  That sucks.

### DIFF
--- a/src/account-settings/reducers.js
+++ b/src/account-settings/reducers.js
@@ -5,7 +5,6 @@ import {
   SAVE_SETTINGS,
   FETCH_TIME_ZONES,
   SAVE_PREVIOUS_SITE_LANGUAGE,
-  DISCONNECT_AUTH,
   UPDATE_DRAFT,
   RESET_DRAFTS,
 } from './actions';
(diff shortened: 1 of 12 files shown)


src/account-settings/reducers.js
```diff
@@ -223,6 +180,15 @@ const reducer = (state = defaultState, action) => {
...
 
+    case DISCONNECT_AUTH.BEGIN:
+    case DISCONNECT_AUTH.SUCCESS:
+    case DISCONNECT_AUTH.FAILURE:
+    case DISCONNECT_AUTH.RESET:
+      return {
+        ...state,
+        thirdPartyAuth: thirdPartyAuthReducer(state.thirdPartyAuth, action),
+      };
+
```
abutterworth: Would `combineReducers` work here? https://redux.js.org/api/combinereducers

src/account-settings/third-party-auth/ThirdPartyAuth.jsx
```diff
@@ -124,18 +123,22 @@ ThirdPartyAuth.propTypes = {
     connected: PropTypes.bool,
     id: PropTypes.string,
   })),
-  disconnectingState: PropTypes.objectOf(PropTypes.oneOf([null, 'pending', 'complete', 'error'])),
-  disconnectErrors: PropTypes.objectOf(PropTypes.bool),
+  providerStatuses: PropTypes.objectOf(PropTypes.oneOf([null, 'pending', 'complete', 'error'])),
```
abutterworth: I think it's worth noting that these statuses are only for disconnection. Not sure if it's worth a comment or different name.

src/account-settings/reducers.js
davidjoy: Yes, that's the plan.

src/account-settings/third-party-auth/ThirdPartyAuth.jsx
davidjoy: Ahh, fair.  Hm.  Will rename. 👍

PR #98, opened 2019-06-05
Adding tests for delete account.

davidjoy:
- comprehensive reducer tests
- UI snapshot tests for most (all?) states

### DIFF
--- a/src/account-settings/delete-account/SuccessModal.test.jsx
+++ b/src/account-settings/delete-account/SuccessModal.test.jsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import renderer from 'react-test-renderer';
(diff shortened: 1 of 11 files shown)


src/account-settings/delete-account/SuccessModal.test.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import renderer from 'react-test-renderer';
+import { IntlProvider, injectIntl } from '@edx/frontend-i18n';
+
+// Testing the modals separately, they just clutter up the snapshots if included here.
```
albemarle: I'm confused by the "if included here" part.

davidjoy: Whoops!  That comment was from DeleteAccount.test.jsx.  Good catch - doesn't belong here. :) . Copy/paste error.

PR #118, opened 2019-10-21
Convert to using frontend-build.

davidjoy:
Goodbye dependencies!  Goodbye config!

I've run all the scripts in the package.json "scripts" section.  Seem good!  Can make a prod build, development server, test, lint, etc.


.env.development
```diff
@@ -0,0 +1,33 @@
...
+FACEBOOK_URL='https://www.facebook.com'
+GOOGLE_PLAY_URL='https://play.google.com/store'
+LINKED_IN_URL='https://www.linkedin.com'
+OPEN_SOURCE_URL='http://localhost:18000/openedx'
+PRIVACY_POLICY_URL='http://localhost:18000/privacy-policy'
+REDDIT_URL='https://www.reddit.com'
+SUPPORT_URL='http://localhost:18000/support'
+TERMS_OF_SERVICE_URL='http://localhost:18000/terms-of-service'
+TWITTER_URL='https://twitter.com'
+YOU_TUBE_URL='https://www.youtube.com'
```
robrap: What changed regarding YouTube?  Are we leaving this in and it just will no longer be used (for now)?

.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+const { createConfig } = require('@edx/frontend-build');
+
+module.exports = createConfig('eslint');
```
robrap: Can you have your editor add new-line to end of file on save?

.npmignore
```diff
@@ -1,13 +1,15 @@
...
-Dockerfile
+babel.config.js
 Makefile
-npm-debug.log
-
-webpack
-coverage
-node_modules
-public
+renovate.json
```
robrap: Same.

package.json
```diff
@@ -1,151 +1,92 @@
...
-    "precommit": "npm run lint",
-    "start": "NODE_ENV=development BABEL_ENV=development webpack-dev-server --config=webpack/webpack.dev.config.js --progress",
-    "test": "jest --coverage --passWithNoTests"
+    "lint": "fedx-scripts eslint",
+    "snapshot": "fedx-scripts jest --updateSnapshot",
+    "start": "fedx-scripts webpack-dev-server --progress",
+    "test": "fedx-scripts jest --coverage --passWithNoTests"
+  },
+  "bugs": {
+    "url": "https://github.com/edx/frontend-app-account/issues"
```
robrap: I thought we don't want to use this for our repos, but maybe I just can't keep up?  We should document this as a decision.  Is it a per-repo decision?  Is there a default decision?

package.json
```diff
@@ -1,151 +1,92 @@
...
-    "react-transition-group": "^2.5.3",
-    "redux": "^4.0.1",
-    "redux-devtools-extension": "^2.13.2",
-    "redux-logger": "^3.0.6",
-    "redux-saga": "^1.0.1",
-    "redux-thunk": "^2.2.0",
-    "reselect": "^4.0.0",
-    "universal-cookie": "^3.1.0",
-    "url-polyfill": "^1.1.5"
+    "@edx/edx-bootstrap": "2.2.1",
```
robrap: We don't want pinning versions to be more meaningful?  In other words, how do I know what is pinned because it must be pinned?  Since we can't add comments here, how do we note this?  In Python, we have a process for how to comment why dependencies need to be pinned (hopefully temporarily).

src/account-settings/reducers.js
```diff
@@ -151,8 +151,8 @@ const reducer = (state = defaultState, action) => {
         countryTimeZones: action.payload.timeZones,
       };
 
-    // TODO: Once all the above cases have been converted into sub-reducers, we can use
-    // combineReducers in this file to greatly simplify it.
+      // TODO: Once all the above cases have been converted into sub-reducers, we can use
+      // combineReducers in this file to greatly simplify it.
```
robrap: Seems like the original indent was fine.

package.json
davidjoy: This is what renovate does when you turn it on - I'm just pre-empting it to save a PR later.

package.json
davidjoy: This is the first I've heard of it...

src/account-settings/reducers.js
davidjoy: Linter complained.  ¯\_(ツ)_/¯

.env.development
davidjoy: I believe we removed the usage of this var in site-footer, yes.  That said, it's also deprecated anyway along with most of these others... the new footer (which I'll be putting in in the next PR) will make them obsolete.  

Point being, more cleanup is forthcoming.

.eslintrc.js
davidjoy: Good call - figured I'd had that, but I didn't.  Fixed.

PR #120, opened 2019-10-22
Adding frontend-base

davidjoy:
It'll be easiest to review this commit by commit.  They're pretty orderly.


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -53,11 +54,29 @@ class AccountSettingsPage extends React.Component {
...
+    this.state = {
+      duplicateTpaProvider,
+    };
   }
 
   componentDidMount() {
+    this.props.fetchUserAccount(this.context.authenticatedUser.username);
     this.props.fetchSettings();
     this.props.fetchSiteLanguages();
+    sendTrackingLogEvent('edx.user.settings.viewed', {
```
abutterworth: Is this true? We've just sent the request to fetch the user account so maybe the page is still blank?

src/account-settings/data/service.js
```diff
@@ -1,41 +1,17 @@
+import { App } from '@edx/frontend-base';
 import pick from 'lodash.pick';
 import omit from 'lodash.omit';
 import isEmpty from 'lodash.isempty';
 
-import { applyConfiguration, handleRequestError, unpackFieldErrors } from '../common/serviceUtils';
-import { configureService as configureDeleteAccountApiService } from './delete-account';
-import { configureService as configureResetPasswordApiService } from './reset-password';
-import { configureService as configureSiteLanguageApiService } from './site-language';
-import { configureService as configureThirdPartyAuthApiService, getThirdPartyAuthProviders } from './third-party-auth';
```
abutterworth: Great to remove so many of these.

src/account-settings/delete-account/data/sagas.js
```diff
@@ -7,7 +7,7 @@ import {
   deleteAccountFailure,
 } from './actions';
 
-import { postDeleteAccount } from './service';
+import postDeleteAccount from './service';
```
abutterworth: I think we should ignore the prefer default export rule here and keep the named export. `postDeleteAccount` from `service.js` seems odd at first read.

src/account-settings/AccountSettingsPage.jsx
davidjoy: I moved this code from index.jsx, where it was being done even earlier.  So this seems more true?  I'm a bit disinclined to change the behavior too much as part of this PR, fwiw.

src/account-settings/data/service.js
davidjoy: Yeah, that was satisfying. :)

src/account-settings/delete-account/data/sagas.js
davidjoy: Yeah... I think you're right.  Going to fix that.

PR #121, opened 2019-10-23
clean up readme

robrap:
@davidjoy: I don't know if you want this in other repos as well?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,8 +5,6 @@ frontend-app-account
 
 This is a micro-frontend application responsible for the display and updating of a user's account information.  Please tag **@edx/arch-team** on any PRs or issues.
 
-----------
-
 Development
 -----------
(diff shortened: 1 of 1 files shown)


davidjoy:
If it's breaking the RST, I suppose we should?

robrap:
@davidjoy: I don't know if it is breaking anything (could be), but it was confusing.

A header could be:
```
----------------------------
Configuration and Deployment
----------------------------
```
Or simpler:
```
Configuration and Deployment
----------------------------
```

I think the top lines (which weren't needed anyway), just moved away from the headers and didn't match length of anything.

I would just clean-up any template level repos which would automatically copy this rst.  I don't know that it is important everywhere.  I did it here, because I thought I was going to update the README, but decided to just commit the clean-up.

PR #124, opened 2019-11-13
Small refactor to make updating the site language sequential

davidjoy:
We were calling the setlang and preferences endpoints in parallel - it’s not clear whether this might create a race condition.  Instead, we now call preferences first, followed by i18n/setlang.  This matches the behavior of the django-based account settings page.

### DIFF
--- a/src/account-settings/data/sagas.js
+++ b/src/account-settings/data/sagas.js
@@ -71,10 +71,13 @@ export function* handleSaveSettings(action) {
     let savedValues = null;
     if (formId === 'siteLanguage') {
       const previousSiteLanguage = yield select(siteLanguageSelector);
-      yield all([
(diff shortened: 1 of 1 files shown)


src/account-settings/data/sagas.js
```diff
@@ -71,10 +71,10 @@ export function* handleSaveSettings(action) {
     let savedValues = null;
     if (formId === 'siteLanguage') {
       const previousSiteLanguage = yield select(siteLanguageSelector);
-      yield all([
-        call(patchPreferences, username, { prefLang: commitValues }),
-        call(postSetLang, commitValues),
-      ]);
+
+      yield call(patchPreferences, username, { prefLang: commitValues });
```
robrap: I'd add a comment about this so that it doesn't change down the line and cause a lot of other headaches.

PR #125, opened 2019-11-20
fix: all account settings page elements now re-localize when the locale changes

davidjoy:
- Ceased using i18nReducer and setLocale from frontend-i18n.
- Caused a number of dropdown menus to re-localize properly.
- Published LOCALE_CHANGED event when changing locale so that AppProvider updates the AppContext accordingly.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -1,4 +1,4 @@
-import { AppContext, fetchUserAccount, App } from '@edx/frontend-base';
+import { AppContext, App } from '@edx/frontend-base';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
@@ -9,6 +9,8 @@ import {
   injectIntl,
   intlShape,
   FormattedMessage,
+  getCountryList,
+  getLanguageList,
 } from '@edx/frontend-i18n';
 import { Hyperlink } from '@edx/paragon';
 
@@ -32,28 +34,8 @@ import {
 import { fetchSiteLanguages } from './site-language';
 
 class AccountSettingsPage extends React.Component {
-  constructor(props) {
-    super(props);
-    this.educationLevels = EDUCATION_LEVELS.map(key => ({
-      value: key,
-      label: props.intl.formatMessage(messages[`account.settings.field.education.levels.${key || 'empty'}`]),
-    }));
-    this.genderOptions = GENDER_OPTIONS.map(key => ({
-      value: key,
-      label: props.intl.formatMessage(messages[`account.settings.field.gender.options.${key || 'empty'}`]),
-    }));
-    this.languageProficiencyOptions = [{
-      value: '',
-      label: props.intl.formatMessage(messages['account.settings.field.language_proficiencies.options.empty']),
-    }].concat(props.languageProficiencyOptions);
-    this.yearOfBirthOptions = [{
-      value: '',
-      label: props.intl.formatMessage(messages['account.settings.field.year_of_birth.options.empty']),
-    }].concat(YEAR_OF_BIRTH_OPTIONS);
-    this.countryOptions = [{
-      value: '',
-      label: props.intl.formatMessage(messages['account.settings.field.country.options.empty']),
-    }].concat(props.countryOptions);
+  constructor(props, context) {
+    super(props, context);
 
     // If there is a "duplicate_provider" query parameter, that's the backend's
     // way of telling us that the provider account the user tried to link is already linked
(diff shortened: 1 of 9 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -32,28 +34,8 @@ import {
 import { fetchSiteLanguages } from './site-language';
 
 class AccountSettingsPage extends React.Component {
-  constructor(props) {
```
davidjoy: Instead of creating these Options lists in the constructor, I memoized them and have them generated on render.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -97,6 +80,29 @@ class AccountSettingsPage extends React.Component {
     return concatTimeZoneOptions;
   });
 
+  getLocalizedOptions = memoize(locale => ({
```
davidjoy: I did them all in one function for convenience - there was only one usage of each of them, so it seemed fine for now.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -354,7 +370,7 @@ class AccountSettingsPage extends React.Component {
             name="siteLanguage"
             type="select"
             options={this.props.siteLanguageOptions}
-            value={this.props.siteLanguage.draftOrSavedValue}
+            value={this.props.siteLanguage.draft !== undefined ? this.props.siteLanguage.draft : this.context.locale}
```
davidjoy: This line replaces part of the need for localeSelector in the selectors.js file.  If we do this conditional here, we can use context instead of redux.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -464,21 +480,12 @@ AccountSettingsPage.propTypes = {
   }).isRequired,
   siteLanguage: PropTypes.shape({
     previousValue: PropTypes.string,
-    draftOrSavedValue: PropTypes.string,
-    savedValue: PropTypes.string,
+    draft: PropTypes.string,
```
davidjoy: This change is the result of getting rid of the use of localeSelector - logic was moved into the conditional above this.  We also no longer need savedValue, as the savedValue is always the same as the context.locale and getLocale()

src/account-settings/AccountSettingsPage.jsx
abutterworth: great!

src/account-settings/data/sagas.js
```diff
@@ -70,16 +69,16 @@ export function* handleSaveSettings(action) {
...
-      const previousSiteLanguage = yield select(siteLanguageSelector);
-
+      const previousSiteLanguage = getLocale();
       // The following two requests need to be done sequentially, with patching preferences before
       // the post to setlang.  They used to be done in parallel, but this might create ambiguous
       // behavior.
       yield call(patchPreferences, username, { prefLang: commitValues });
       yield call(postSetLang, commitValues);
 
-      yield put(setLocale(commitValues));
```
davidjoy: No need to setLocale (for redux) anymore!

src/account-settings/AccountSettingsPage.jsx
abutterworth: LGTM – they all change with locale anyway

src/account-settings/data/sagas.js
```diff
@@ -70,16 +69,16 @@ export function* handleSaveSettings(action) {
...
-      yield put(savePreviousSiteLanguage(previousSiteLanguage.savedValue));
+      yield put(savePreviousSiteLanguage(previousSiteLanguage));
+
+      App.publish(LOCALE_CHANGED, getLocale());
```
davidjoy: frontend-i18n doesn't publish this itself when the locale changes (and we don't have a way of telling it the locale has changed) - so we do it manually here instead.

src/account-settings/data/selectors.js
```diff
@@ -123,40 +107,24 @@ const activeAccountSelector = createSelector(
...
- * This selector converts the site language code back to the server version so that it can match up
- * with one of the options in the site language dropdown.  The drafts version will already be the
- * server version, but if it's from localeSelector, it will be our client (two character) version.
- */
 export const siteLanguageSelector = createSelector(
   previousSiteLanguageSelector,
   draftsSelector,
-  localeSelector,
-  (previousValue, drafts, locale) => ({
+  (previousValue, drafts) => ({
```
davidjoy: Removing the need for localeSelector

src/account-settings/AccountSettingsPage.jsx
davidjoy: Exactly.

davidjoy: They just don't all use it directly. :)

PR #192, opened 2020-03-04
Add phone number field and coaching consent toggle

Tj-Tracy:
MB-196

Adds phone number and coaching consent toggle to the profile. Currently,
toggled off in production, until we are ready for MB learners to start
recieving coaching. Screenshots:
___
Screen Shot 2020-03-04 at 4 42 45 PM

___

image

___

Screen Shot 2020-03-04 at 4 43 25 PM


src/account-settings/_style.scss
```diff
@@ -35,4 +35,13 @@
     margin-bottom: map-get($spacers, 5);
     padding-top: 1rem;
   }
+  
+  .custom-switch {
```
justinhynes: Nit: I don't know if we have a naming convention but this name seems really... generic? Something like 'custom-consent-switch' seems to give it a bit more meaning as to what this styling belongs to?

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -327,6 +327,45 @@ class AccountSettingsPage extends React.Component {
             emptyLabel={this.props.intl.formatMessage(messages['account.settings.field.language.proficiencies.empty'])}
             {...editableFieldProps}
           />
+          {process.env.TEMP_COACHING_FEATURE_FLAG &&
+            <>
```
justinhynes: Noob question: This may be better for an offline/Slack message... but I'm curious what the '<>' is for? Why is it needed?

src/account-settings/_style.scss
Tj-Tracy: its a bootstrap util class that I wanted to override _only_ when its on this page. Hence the generic name.

src/account-settings/AccountSettingsPage.jsx
Tj-Tracy: shorthand for `<React.Fragment />`. Its need because the render needs to return only one child element after the `&&`. The point of the fragment is to encapsulate it in a way that doesn't change the actual markup on the page (React fragments don't render to the DOM).

src/account-settings/_style.scss
justinhynes: 👍

src/account-settings/AccountSettingsPage.jsx
justinhynes: Thanks 👍

src/account-settings/data/service.js
```diff
@@ -149,17 +149,46 @@ export async function getProfileDataManager(username, userRoles) {
...
+ * @param {Number} userId users are identified in the api by LMS id
+ * @param {Object} commitValues { coaching_consent }
+ */
+export async function patchCoachingPreferences(userId, commitValues) {
+  const requestUrl = `${getConfig().LMS_BASE_URL}/api/coaching/v1/users/${userId}/`;
+  const body = {
+    ...commitValues,
+    user: userId,
+  };
+  const options = { headers: { 'Content-Type': 'application/json' } };
```
davidjoy: Is adding this header necessary?  I seem to recall seeing it in a request where it wasn't explicitly set... 🤔

src/account-settings/data/service.js
```diff
@@ -168,20 +197,23 @@ export async function getSettings(username, userRoles) {
     thirdPartyAuthProviders: results[2],
     profileDataManager: results[3],
     timeZones: results[4],
+    coaching_consent: results[5],
```
davidjoy: As a rule of thumb, we've been trying to convert from snake_case to camelCase (and back again) in our API layer so that the rest of the app code can be consistent - it can be confusing and error prone otherwise, and makes it hard to find-and-replace and/or see all the places something is used.

.env.development
```diff
@@ -16,3 +16,5 @@ SEGMENT_KEY=null
 SITE_NAME='edX'
 SUPPORT_URL='http://localhost:18000/support'
 USER_INFO_COOKIE_NAME='edx-user-info'
+# Temporary, Remove this once we are ready to release the feature.
+COACHING_ENABLED=true
```
davidjoy: Think we may want to leave this `false` by default.  Also, it may need to be added to the other `.env` files or it may not work in prod/testing.

package.json
```diff
@@ -29,22 +29,23 @@
...
-    "@fortawesome/react-fontawesome": "0.1.8",
+    "@fortawesome/react-fontawesome": "0.1.9",
     "babel-polyfill": "6.26.0",
     "classnames": "2.2.6",
     "font-awesome": "4.7.0",
     "form-urlencoded": "4.0.1",
     "formdata-polyfill": "3.0.19",
     "history": "4.10.1",
     "lodash.camelcase": "4.3.0",
+    "lodash.debounce": "^4.0.8",
```
davidjoy: Renovate will complain about this `^`. :)

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -327,6 +328,14 @@ class AccountSettingsPage extends React.Component {
             emptyLabel={this.props.intl.formatMessage(messages['account.settings.field.language.proficiencies.empty'])}
             {...editableFieldProps}
           />
+          {getConfig().COACHING_ENABLED &&
```
davidjoy: Nice!

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -474,10 +483,17 @@ AccountSettingsPage.propTypes = {
...
+    coaching_consent: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
     social_link_linkedin: PropTypes.string,
     social_link_facebook: PropTypes.string,
     social_link_twitter: PropTypes.string,
     time_zone: PropTypes.string,
+    coaching: PropTypes.objectOf(PropTypes.shape({
+      coaching_consent: PropTypes.string.isRequired,
+      user: PropTypes.number.isRequired,
+      eligible_for_coaching: PropTypes.bool.isRequired,
+    })).isRequired,
```
davidjoy: Will `coaching` having `isRequired` be a problem if it's turned off?

src/account-settings/coaching/CoachingToggle.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+CoatchingToggle.defaultProps = {
+  phone_number: '',
+  error: '',
+};
+
+CoatchingToggle.propTypes = {
+  name: PropTypes.string.isRequired,
+  error: PropTypes.string,
+  coaching: PropTypes.objectOf(PropTypes.shape({
```
davidjoy: Regarding my comment on the `coaching` prop earlier - it could be required here and not required in the AccountSettingsPage, I think, and we'd still catch mistakes if it were somehow not defined.

.env.development
Tj-Tracy: Turned off and added to all `.env`s for consistency.

package.json
Tj-Tracy: 🛠 'd

src/account-settings/AccountSettingsPage.jsx
Tj-Tracy: I've made this one no longer required.

src/account-settings/coaching/CoachingToggle.jsx
Tj-Tracy: I've kept this one as required.

PR #195, opened 2020-03-16
Fix 404 error for coaching

Tj-Tracy:


.env.development
```diff
@@ -17,4 +17,4 @@ SITE_NAME='edX'
 SUPPORT_URL='http://localhost:18000/support'
 USER_INFO_COOKIE_NAME='edx-user-info'
 # Temporary, Remove this once we are ready to release the feature.
-COACHING_ENABLED=false
+COACHING_ENABLED=''
```
staubina: Why not default to false?

Tj-Tracy: false gets read in as a string, not the value. a weird quirk of `dotev`

PR #198, opened 2020-03-19
Fix timezone dropdown under Site Preferences

adeelkhan:
This patch would fix timezone dropdown which
doesn't change  value on selection.

PROD-1349


adeelkhan:
@edx/arch-team Please review.

PR #205, opened 2020-03-25
Add coaching consent form

tuchfarber:
Screenshot_2020-03-31 Account edX
Screenshot_2020-03-31 Account edX(1)
Screenshot_2020-03-31 Account edX(2)

Add coaching consent form.


tuchfarber:
Things I'm mostly worried about:
 - Having to manage the redux state like I am. It was the only way I could think of doing it without mucking about in the codebase too much. I have a long comment on the reasoning under the componentDidUpdate function
- The renderView design I made. I know we usually do {something ? (thing_1) : (thing_2)}, but with 6 states and `else if`s determining state, the template got really gross. If there's a cleaner way to handle this, lemme know.

src/account-settings/coaching/CoachingConsent.jsx
```diff
@@ -0,0 +1,293 @@
...
+};
+
+class CoachingConsent extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+
+    // Used to redirect back to the courseware.
+    const nextUrl = this.sanitizeForwardingUrl(getQueryParameters().next);
+    this.state = {
+      redirectUrl: nextUrl || '/',
```
Tj-Tracy, line 56: This will go back to account settings right? should we redirect back to dashboard by default?

src/index.jsx
```diff
@@ -11,23 +11,33 @@ import Footer, { messages as footerMessages } from '@edx/frontend-component-foot
...
+import CoachingConsent from './account-settings/coaching/CoachingConsent';
 import appMessages from './i18n';
 
 import './index.scss';
 import './assets/favicon.ico';
 
+const HeaderFooterLayout = ({ children }) => (
+  <div>
+    <Header />
+      <main>
```
Tj-Tracy, line 23: possible a11y concern: We will most likely want to wrap the new route in a main tag as well.

src/account-settings/coaching/CoachingConsent.jsx
```diff
@@ -0,0 +1,293 @@
...
+      if (this.state.allSubmissionsComplete) {
+        currentView = VIEWS.SUCCESS;
+      } else {
+        currentView = VIEWS.SUCCESS_PENDING;
+      }
+    } else if (this.state.declineSubmitted && !formHasErrors) {
+      if (this.state.allSubmissionsComplete) {
+        currentView = VIEWS.DECLINED;
+      } else {
+        currentView = VIEWS.DECLINE_PENDING;
```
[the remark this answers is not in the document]
Tj-Tracy: I think this structure works out fine. the only change I might consider is converting this into a switch statement as well.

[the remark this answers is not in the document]
tuchfarber: They were different states so I could have different `srMessages` on the loader. Also not sure how I'd change this to a switch statement? It has nested `if`s and switch statements match a single expression to a series of a values. I need to check multiple expressions here (which is why I used both).

src/account-settings/coaching/CoachingConsent.jsx
tuchfarber: Dashboard is a better option. Will fix.

src/index.jsx
tuchfarber: Will fix

src/account-settings/data/selectors.js
```diff
@@ -159,3 +169,32 @@ export const accountSettingsPageSelector = createSelector(
     tpaProviders: accountSettings.thirdPartyAuth.providers,
   }),
 );
+
+export const coachingConsentPageSelector = createSelector(
+  accountSettingsSelector,
+  formValuesSelector,
+  hiddenFieldsSelector,
```
zamanafzal, line 176: @tuchfarber I am working with the selector(hiddenFieldsSelector) and going to remove this method in my PR. Do you think you need it in the coachingConsentPageSelector?

tuchfarber: Nope, that's fine to delete. Must have missed it when I was cleaning up. Thanks for asking!

PR #206, opened 2020-04-01
ENT-2651 Recovery email Field UX logic on Account Settings page was not same to Dashboard

zamanafzal:


src/account-settings/data/service.js
```diff
@@ -136,14 +136,9 @@ export async function getProfileDataManager(username, userRoles) {
   if (userRoleNames.includes('enterprise_learner')) {
     const url = `${getConfig().LMS_BASE_URL}/enterprise/api/v1/enterprise-learner/?username=${username}`;
     const { data } = await getAuthenticatedHttpClient().get(url).catch(handleRequestError);
-
     if ('results' in data) {
-      for (let i = 0; i < data.results.length; i += 1) {
-        const enterprise = data.results[i].enterprise_customer;
-        if (enterprise.sync_learner_profile_data) {
```
moconnell1453, line 143: I don't think we should change the logic here. getProfileDataManager() is used elsewhere in this MFE, where the sync_learner_profile_data flag is relevant. It would be preferable to have a similar method that ignores this flag that we can use for the secondary email check. Or, add a parameter to this method to ignore the flag.

zamanafzal: Yes, you are right. This change is related to the other bug of wrong enterprise name showing on the Account Settings Page.

zamanafzal: @moconnell1453 I have addressed your feedback. Can you please take another look.

zamanafzal:
@tuchfarber I have removed hiddenFieldSelector from your code. Can you please verify.

tuchfarber:
Looks good!

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -184,7 +184,7 @@ class AccountSettingsPage extends React.Component {
   }
 
   renderSecondaryEmailField(editableFieldProps) {
-    if (this.props.hiddenFields.includes('secondary_email')) {
+    if (!Boolean(this.props.formValues.secondary_email_enabled)) {
```
muhammad-ammar, line 187: Any reason for using `Boolean` here?

zamanafzal: Yeah, to make sure it returns Bool value, as `secondary_email_enabled ` can be null at sometimes.

PR #210, opened 2020-04-16
Hackathon XXIV -- Add ID verification page

kdmccormick:
For Hackathon XXIV IDv page overhaul.

See https://github.com/edx/edx-platform/pull/23760
Also https://openedx.atlassian.net/wiki/spaces/HACK/pages/1371242822/Hackathon+XXIV+IDv+Overhaul

Note from George Babey: We need to consider scenarios where the account name change fails. Specifically, learners who SSO via IdP may have a locked-down profile.


bseverino:
@edx/masters-devs-cosmonauts This PR is ready for review

src/id-verification/routing-utilities.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const nextIndex = panelSteps.indexOf(originSlug) + 1;
+  return nextIndex < panelSteps.length ? panelSteps[nextIndex] : null;
+};
+
+// check if the user is too far into the flow and if so, return the slug of the
+// furthest panel they are allow to be.
+export const useVerificationRedirectSlug = (slug) => {
+  // TODO: remove this short-circuit after development is done
```
Zacharis278, line 33: Can this be removed and does this work properly if it is?

src/id-verification/data/service.js
```diff
@@ -0,0 +1,71 @@
...
+ *   - courseRunKey (String|null): Optional course run to redirect to.
+ *
+ * Returns { success: Boolean, message: String|null }
+ */
+export async function submitIdVerfication(verificationData) {
+  const keyMap = {
+    facePhotoFile: 'face_image',
+    idPhotoFile: 'photo_id_image',
+    idPhotoName: 'full_name',
+    courseRunKey: 'course_id',
```
Zacharis278, line 48: Looks to me that we're not entirely supporting a redirect back to the course where the verification link was clicked.  IMO we don't need it to merge this but let's definitely open a ticket to add that feature at some point.

src/id-verification/routing-utilities.js
bseverino: It works. However I did notice an issue where if the user reloads during the process, camera access is lost and they're prompted to upload an image instead. Added a redirect to the beginning of the process on render to avoid this, I'll push up the change soon.

src/id-verification/IdVerification.messages.js
```diff
@@ -0,0 +1,346 @@
...
+    id: 'id.verification.submitted.text',
+    defaultMessage: 'We have received your information and are verifying your identity. You will see a message on your dashboard when the verification process is complete (usually within 1-2 days). In the meantime, you can still access all available course content.',
+    description: 'Text confirming that ID verification request has been received.',
+  },
+  'id.verification.return': {
+    id: 'id.verification.submitted.return',
+    defaultMessage: 'Return to Your Dashboard',
+    description: 'Button to return to the dashboard.',
+  },
+});
```
kdmccormick, line 344: Thanks for writing great translator descriptions for all of these messages!

src/id-verification/IdVerificationContext.jsx
```diff
@@ -0,0 +1,93 @@
...
+      {children}
+    </IdVerificationContext.Provider>
+  );
+}
+IdVerificationContextProvider.propTypes = {
+  children: PropTypes.node,
+};
+IdVerificationContextProvider.defaultProps = {
+  children: undefined,
+};
```
kdmccormick, line 87:
```suggestion
IdVerificationContextProvider.propTypes = {
  children: PropTypes.node.isRequired,
};
```
If the default value is `undefined`, should/can we make the prop required?

src/id-verification/ImageFileUpload.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+
+export default function ImageFileUpload({ onFileChange }) {
+  const handleChange = useCallback((e) => {
+    if (e.target.files.length === 0) {
+      // do something else
+      return;
+    }
```
kdmccormick, line 9: Looks like a left-over TODO comment. Do we need to handle this condition any differently, or can we just delete the comment?

src/id-verification/data/service.js
kdmccormick: @bseverino I recommend mentioning that the redirect doesn't work in a comment, referencing JIRA ticket number for the fix.

src/id-verification/ImageFileUpload.jsx
bseverino: I don't think anything else needs to be done there; I'll go ahead and delete the comment.

bseverino:
Redirect to course currently does not work and should be added in the future. More info: [MST-282](https://openedx.atlassian.net/browse/MST-282)

kdmccormick:
@bseverino Whoops, I meant an in-code comment for MST-282. My bad for the ambiguity!

src/id-verification/index.js
```diff
@@ -0,0 +1,4 @@
+export { default } from './IdVerificationPage';
+// export { default as reducer } from './data/reducers';
+// export { default as saga } from './data/sagas';
+// export { storeName } from './data/selectors';
```
kdmccormick, line 4: Feel free to drop these -- they were the result of copypasta.

kdmccormick:
@davidjoy or @abutterworth  -- We'd like to merge this. Users won't see it unless they guess the URL.

Would you mind approving this PR and/or giving us write access to the repository? @bseverino will be opening more ID-verification-related PRs throughout the summer.

PR #211, opened 2020-04-21
ENT-2741: Fixed Profile Manager Banner Scenario

sameenfatima78:
Profile manager banner should only be visible if:

- a learner is linked to an enterprise
- that enterprise manages the settings

Also, it should display the name of the active enterprise in case if a learner is linked to multiple enterprises. 

**Jira**: https://openedx.atlassian.net/browse/ENT-2741

### DIFF
--- a/src/account-settings/data/service.js
+++ b/src/account-settings/data/service.js
@@ -138,11 +138,10 @@ export async function getProfileDataManager(username, userRoles) {
     const { data } = await getAuthenticatedHttpClient().get(url).catch(handleRequestError);
 
     if ('results' in data) {
-      for (let i = 0; i < data.results.length; i += 1) {
-        const enterprise = data.results[i].enterprise_customer;
-        if (enterprise.sync_learner_profile_data) {
-          return enterprise.name;
-        }
+      const enterprise = data.results[0].enterprise_customer;
+      // To ensure that enterprise returned is current enterprise & it manages profile settings
+      if (enterprise.sync_learner_profile_data) {
+        return enterprise.name;
       }
     }
   }


src/account-settings/data/service.js
```diff
@@ -140,7 +140,8 @@ export async function getProfileDataManager(username, userRoles) {
     if ('results' in data) {
       for (let i = 0; i < data.results.length; i += 1) {
         const enterprise = data.results[i].enterprise_customer;
-        if (enterprise.sync_learner_profile_data) {
+        // Return the active enterprise in case a learner is linked to multiple enterprises
+        if (data.results[i].active) {
```
zamanafzal, line 144: Also, I think at the place of `data.results[i].active` we can use` enterprise.active`. One more thing that you can do is return the first enterprise which will be returned from the API call. As the user active enterprise will always be on the zero index.

sameenfatima78: I think the `active` flag in `enterprise_customer` is used for other purpose. The active flag in results is the one that shows the currently active enterprise in session. I think returning the first enterprise is a better approach as it always returns the active enterprise first.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -145,7 +145,7 @@ class AccountSettingsPage extends React.Component {
   }
 
   renderManagedProfileMessage() {
-    if (!this.isManagedProfile()) {
+    if (!this.isManagedProfile() || !Boolean(this.props.formValues.secondary_email_enabled)) {
```
muhammad-ammar, line 148: @sameenfatima78 @zamanafzal Please help me understand why banner display is dependent on `secondary email feature`? What is connection between `secondary email feature` and `banner display`?

sameenfatima78: I also raised the same question yesterday and according to @zamanafzal there's no relationship between the two. It's just a feature which will be enabled only if the `enable_secondary_email_feature` switch is enabled. This is being done to make the MFE logic consistent. https://openedx.atlassian.net/browse/ENT-2741

src/account-settings/data/service.js
```diff
@@ -140,7 +140,8 @@ export async function getProfileDataManager(username, userRoles) {
     if ('results' in data) {
       for (let i = 0; i < data.results.length; i += 1) {
         const enterprise = data.results[i].enterprise_customer;
-        if (enterprise.sync_learner_profile_data) {
```
sameenfatima78, line 143: @moconnell1453 I read your comment on @zamanafzal [PR](https://github.com/edx/frontend-app-account/pull/206#discussion_r402315677) that we shouldn't change this logic as 
`getProfileDataManager()` is being used elsewhere. Can you please help me identify where else is it being called from? I searched it in the repo but couldn't find any other callers. I'm trying to use the ignore flag approach as you suggested in the comment and want to know from where should I pass the flag value so the flow is not affected for the other callers.

moconnell1453: getProfileDataManager() is only called once, but that is in getSettings(), and the result of this call is returned as part of the settings, and that particular setting is referenced in several places as the profileDataManager property: in isManagedProfile(), renderEmptyStaticMessage() and renderManagedProfileMessage().

Regardless, we don't want to remove the check on sync_learner_profile_data because that flag indicates that, not only is the user a member of an enterprise, but that their settings are managed by that enterprise. So, in the context of this story, we need to check that flag, otherwise we don't know if the enterprise is managing the settings or not. In my comments for Zaman's PR, I wanted us to not check that flag for the purposes of determining whether or not the recovery email field should be shown because recovery email functionality should be independent of the management status of the enterprise.

sameenfatima78: Got it, thanks for clarifying. This makes sense.

sameenfatima78: @moconnell1453 Also, wouldn't it be better if we remove the dependency of `enable_secondary_email_feature` switch for Enterprise Name banner? I think the enterprise name banner functionality should be independent of it.  @muhammad-ammar also raised the same question [yesterday](https://github.com/edx/frontend-app-account/pull/211/files#r412777391)

src/account-settings/AccountSettingsPage.jsx
moconnell1453: I agree with removing this flag check, though you are right if we wanted to preserve the old functionality. I believe that the original intention was that recovery email would only be an option for learner's whose settings are managed by their enterprise. The recovery email would allow them to maintain access to their account if they ever left that enterprise.

However, having reviewed this, we agreed that it doesn't make much sense to limit the use of recovery email for only that purpose. In fact, the Sustainability team is expanding recovery email usage to B2C customers, so this flag check would likely get dropped anyway.

src/account-settings/data/service.js
```diff
@@ -138,11 +138,10 @@ export async function getProfileDataManager(username, userRoles) {
     const { data } = await getAuthenticatedHttpClient().get(url).catch(handleRequestError);
 
     if ('results' in data) {
-      for (let i = 0; i < data.results.length; i += 1) {
-        const enterprise = data.results[i].enterprise_customer;
-        if (enterprise.sync_learner_profile_data) {
-          return enterprise.name;
-        }
+      const enterprise = data.results[0].enterprise_customer;
```
moconnell1453, line 141: In this case, we're just choosing the first result returned. Do we know if that is the correct enterprise? By that I mean, is it the enterprise that the learner is currently logged in to?

sameenfatima78: Yes, the API call always returns the "active" enterprise at 0th index. @zamanafzal  also confirmed and in fact suggested this change.

src/account-settings/AccountSettingsPage.jsx
sameenfatima78: Okay, I'll remove this flag check and update the PR shortly. Thank you for your feedback.

src/account-settings/data/service.js
moconnell1453: Okay. I see his earlier comment now. Thank you!

PR #218, opened 2020-05-01
Reset coaching submission state on submit

tuchfarber:
Prior to this, if you attempted to sign up for coaching and failed, then went to decline, you would be sent to the success screen instead of redirected. This also removes the default from the coaching API check now that it won't 404 when the user requests their own data.

### DIFF
--- a/src/account-settings/coaching/data/service.js
+++ b/src/account-settings/coaching/data/service.js
@@ -8,20 +8,8 @@ import get from 'lodash.get';
  * @param {Number} userId users are identified in the api by LMS id
  */
 export async function getCoachingPreferences(userId) {
-  let data = null;
-  try {
-    ({ data } = await getAuthenticatedHttpClient()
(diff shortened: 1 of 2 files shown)


src/account-settings/coaching/data/service.js
```diff
@@ -8,20 +8,8 @@ import get from 'lodash.get';
...
-    // UserCoaching model yet.
-    data = {
-      coaching_consent: false,
-      user: userId,
-      eligible_for_coaching: false,
-      consent_form_seen: false,
-    };
-  }
+  const { data } = await getAuthenticatedHttpClient()
+    .get(`${getConfig().LMS_BASE_URL}/api/coaching/v1/users/${userId}/`);
```
staubina, line 12: The PR looks good, any reason we removed the error handling here?

staubina: I read the notes again and it makes more sense now 👍

PR #220, opened 2020-05-01
MICROBA-311 Add state field to account settings

oliviaruizknott:
**DO NOT MERGE. [edx-plaform#23877](https://github.com/edx/edx-platform/pull/23877) must be merged first.**

Add state field dropdown to account settings page if country is US;
add states list; handle localization

MICROBA-311
https://openedx.atlassian.net/browse/MICROBA-311

Functionality can be tested with code in this `edx/edx-plaform` PR: [MICROBA-311 Add US State field to UserProfile](https://github.com/edx/edx-platform/pull/23877). (The branch to checkout is `ork/MICROBA-311_add-state-to-user-profile`.)

A user should be able to add a US State in their account profile if they have selected the United States as their country. Data in the state field should be cleared if the user then changes their country to something other than the United States. 

**See functionality working locally here:**
https://www.loom.com/share/13bf4b39d704453694e70414b47f66b0

**Note**: For some reason, the screen capture can’t see the default drop-down select menus. They definitely work, but just aren’t showing up in the recording.

You should see:
- Adding the United States as a country makes the state field visible.
- State and country persist on page reload.
- Changing the country to something other than the US removes the state field.
- Changing back to US does not repopulate the state field with the originally selected state (because it was removed on country change).


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -31,6 +31,7 @@ import {
   YEAR_OF_BIRTH_OPTIONS,
   EDUCATION_LEVELS,
   GENDER_OPTIONS,
+  getStatesList,
```
Tj-Tracy, line 34: We should probably move this and the new constants into `frontend-platform/i18n`, as there are most likely going to be other places we may want to use a state list.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -209,11 +214,15 @@ class AccountSettingsPage extends React.Component {
...
+      stateOptions,
       languageProficiencyOptions,
       yearOfBirthOptions,
       educationLevelOptions,
       genderOptions,
-    } = this.getLocalizedOptions(this.context.locale);
+    } = this.getLocalizedOptions(this.context.locale, this.props.formValues.country);
+
+    // Show State field only if the country is US (could include Canada later)
+    const showState = this.props.formValues.country == "US";
```
staubina, line 225: Like the other PR, consider making this into a Constant to help with code readability in the future.

PR #222, opened 2020-05-05
Prevent phone_number save on coaching save failure

Tj-Tracy:
Originally, we had it set up so that the coaching consent form saved 33
values in parallel. This add a new saga (saveMultipleSettings) to save
those 3 values in sync. This way, if one fails, the rest of the calls
fail.

Something to watch out for here: the order matters. The array that you
give the save function must be ordered in the way you need the data
saved.

### DIFF
--- a/src/account-settings/data/sagas.js
+++ b/src/account-settings/data/sagas.js
@@ -12,6 +12,7 @@ import {
   fetchSettingsFailure,
   closeForm,
   SAVE_SETTINGS,
+  SAVE_MULTIPLE_SETTINGS,
   saveSettingsBegin,
   saveSettingsSuccess,
   saveSettingsFailure,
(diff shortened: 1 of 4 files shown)


src/account-settings/data/sagas.js
```diff
@@ -100,6 +104,28 @@ export function* handleSaveSettings(action) {
...
+
+// handles mutiple settings saved at once, in order, and stops executing on first failure.
+export function* handleSaveMultipleSettings(settings) {
+  try {
+    yield put(saveMultipleSettingsBegin());
+    const { username, userId } = getAuthenticatedUser();
+    const { settingsArray } = settings.payload;
+    for (let i = 0; i < settingsArray.length; i += 1) {
+      const { formId, commitValues } = settingsArray[i];
+      yield call(patchSettings, username, { [formId]: commitValues }, userId);
```
tuchfarber, line 116: Does this need to `await` the `async patchSettings`? I'm not sure how `yield` and `call` work with promises.

Tj-Tracy: At risk of talking about things I don't fully understand, `call` will handle that  for us. `sagas` uses generators to manage the async calls under the hood.

PR #230, opened 2020-05-15
MB-366 don't save non eligible coaching number

Tj-Tracy:
This fixes a few little bugs in the system as well as the one listed in
the ticket.

* We no longer make an extra API call to user profile when toggling
coaching
* The user now must toggle coaching off to save a non-eligible phone
number
* If the user changes to another US based phone number, or is not
consented to coaching, the phone number will save as normal.

* SAVE_MULTIPLE_SETTINGS now leverages the SAVE_SETTINGS action for a full redux log and a synchronous execution.

* coaching toggled on by default in dev

### DIFF
--- a/src/account-settings/data/reducers.js
+++ b/src/account-settings/data/reducers.js
@@ -154,7 +154,7 @@ const reducer = (state = defaultState, action) => {
     case SAVE_MULTIPLE_SETTINGS.SUCCESS:
       return {
         ...state,
-        saveState: 'success',
+        saveState: 'complete',
       };
 
     case SAVE_MULTIPLE_SETTINGS.FAILURE:
--- a/src/account-settings/data/sagas.js
+++ b/src/account-settings/data/sagas.js
@@ -115,8 +115,7 @@ export function* handleSaveMultipleSettings(settings) {
       const { formId, commitValues } = settingsArray[i];
       yield put(saveSettingsBegin());
       const commitData = { [formId]: commitValues };
-      let savedSettings = commitData;
-      savedSettings = yield call(patchSettings, username, commitData, userId);
+      const savedSettings = yield call(patchSettings, username, commitData, userId);
       yield put(saveSettingsSuccess(savedSettings, commitData));
     }
     yield put(saveMultipleSettingsSuccess(settings));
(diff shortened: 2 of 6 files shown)


src/account-settings/data/sagas.js
```diff
@@ -113,7 +113,11 @@ export function* handleSaveMultipleSettings(settings) {
     const { settingsArray } = settings.payload;
     for (let i = 0; i < settingsArray.length; i += 1) {
       const { formId, commitValues } = settingsArray[i];
-      yield call(patchSettings, username, { [formId]: commitValues }, userId);
+      yield put(saveSettingsBegin());
+      const commitData = { [formId]: commitValues };
+      let savedSettings = commitData;
+      savedSettings = yield call(patchSettings, username, commitData, userId);
+      yield put(saveSettingsSuccess(savedSettings, commitData));
```
tuchfarber, line 120: I'm confused at what's happening here. We're saving the object we're sending to the API in `savedSettings`, then immediately overwriting it with the response data? What's the benefit of setting it to the `commitData`? (This is probably some generator magic I'm missing).

src/account-settings/data/reducers.js
```diff
@@ -154,7 +154,7 @@ const reducer = (state = defaultState, action) => {
     case SAVE_MULTIPLE_SETTINGS.SUCCESS:
       return {
         ...state,
-        saveState: 'success',
+        saveState: 'complete',
```
tuchfarber, line 157: Just for clarity, was this changed because `SAVE_MULTIPLE_SETTINGS.SUCCESS` only means the API call was successful and not the validation? Seems like it's a no-op since it's changed in both spots, but curious if it's for more than clarity.

Tj-Tracy: Correct. Really the only reason I changed eliminate some warnings. `SAVE_MULTIPLE_SETTINGS` will only `complete` if all of the settings in the array complete without error.

src/account-settings/data/sagas.js
Tj-Tracy: Ahh, thats actually not right. I had copied some code over and missed changing this.

PR #242, opened 2020-06-17
MICROBA-309 Collect demographics on account page

oliviaruizknott:
Adds demographics questions to the account page.

MICROBA-309


src/account-settings/demographics/DropdownCheckbox.jsx
```diff
@@ -0,0 +1,70 @@
+import React, { useEffect, useState } from 'react';
+import { CheckBox } from '@edx/paragon';
+import { DECLINED } from '../data/constants';
+
+export const DropdownCheckbox = (props) => {
```
oliviaruizknott, line 5: 🤔 This is no longer a good name.

src/account-settings/EditableField.jsx
```diff
@@ -106,6 +106,7 @@ function EditableField(props) {
                 options={options}
                 {...others}
               />
+              <>{others.children}</>
```
oliviaruizknott, line 109: This change doesn’t (or shouldn’t and doesn’t seem to be) affect existing fields, but it allows us squeeze in a second / custom input to handle some of the new kinds of inputs we want on this page.

src/account-settings/data/actions.js
```diff
@@ -100,9 +100,9 @@ export const savePreviousSiteLanguage = previousSiteLanguage => ({
   payload: { previousSiteLanguage },
 });
 
-export const saveMultipleSettings = settingsArray => ({
+export const saveMultipleSettings = (settingsArray, formId=null) => ({
   type: SAVE_MULTIPLE_SETTINGS.BASE,
-  payload: { settingsArray },
+  payload: { settingsArray, formId },
```
oliviaruizknott, line 105: @Tj-Tracy, looks like you wrote this stuff recently. Especially hoping for your eyes on this to make sure it’s alright. (I don’t have coaching working locally so haven’t click-tested this yet.)

src/account-settings/data/sagas.js
```diff
@@ -106,19 +106,23 @@ export function* handleSaveSettings(action) {
...
       const commitData = { [formId]: commitValues };
       const savedSettings = yield call(patchSettings, username, commitData, userId);
       yield put(saveSettingsSuccess(savedSettings, commitData));
     }
-    yield put(saveMultipleSettingsSuccess(settings));
+    yield put(saveMultipleSettingsSuccess(action));
+    if (formId) {
+      yield delay(1000);
+      yield put(closeForm(formId));
+    }
```
oliviaruizknott, line 125: @Tj-Tracy All the changes to this `saveMultipleSettings` code is for this: I want to (optionally) be able to close the open setting.

src/account-settings/data/service.js
```diff
@@ -1,12 +1,14 @@
...
 import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import pick from 'lodash.pick';
+import pickBy from 'lodash.pickby';
 import omit from 'lodash.omit';
 import isEmpty from 'lodash.isempty';
 
 import { handleRequestError, unpackFieldErrors } from './utils';
 import { getThirdPartyAuthProviders } from '../third-party-auth';
 import { getCoachingPreferences, patchCoachingPreferences } from '../coaching/data/service';
+import { getDemographics, postDemographics, patchDemographics } from '../demographics/data/service';
```
oliviaruizknott, line 11: Note to self: don’t need `postDemographics` here.

src/account-settings/data/service.js
```diff
@@ -161,6 +163,7 @@ export async function getSettings(username, userRoles, userId) {
     getProfileDataManager(username, userRoles),
     getTimeZones(),
     getConfig().COACHING_ENABLED && getCoachingPreferences(userId),
+    getDemographics(userId),
```
oliviaruizknott, line 166: Need to gate behind feature flag.

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -0,0 +1,264 @@
...
+    }).join(", ")
+  }
+
+  handleEditableFieldChange = (name, value) => {
+    this.props.updateDraft(name, value);
+  };
+
+  handleSubmit = (formId, values) => {
+    // We have some custom fields in this section. Instead of relying on the
+    // submitted values, submit the values stored in 'drafts'.
```
oliviaruizknott, line 85: Significant! The rest of the account page stores things differently than I’m choosing to here. Because of how we’re squeezing in custom inputs in some cases, the `Submit` button doesn’t always know what values it should be including. Instead, this takes all the values in `drafts` (the redux store keeps track of this on any field change) and submits them when the `Submit` button is clicked.

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -0,0 +1,264 @@
...
+              id='field-demographics_gender_description'
+              type='text'
+              placeholder={this.props.intl.formatMessage(messages['account.settings.field.demographics.gender_description.empty'])}
+              value={this.props.formValues.demographics_gender_description}
+              onChange={(e) => this.handleEditableFieldChange(`demographics_gender_description`, e.target.value)}
+              aria-label={this.props.intl.formatMessage(messages['account.settings.field.demographics.gender_description'])}
+              className="mt-1"
+            />
+          }
+        </EditableField>
```
oliviaruizknott, line 145: This `gender` field is one of our custom handlings. We shove in another text input if the user selects `self-describe`.

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -0,0 +1,264 @@
...
+        <EditableField
+          name="demographics_user_ethnicity"
+          type="select"
+          hidden={true}
```
oliviaruizknott, line 149: Another custom handling. `hidden` is passed here because, in the case, we only want the `EditableField` stuff for how it lets us switch between editing and display. But the input itself (custom checkboxes) is being passed in below as a child (not as props).

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -0,0 +1,264 @@
...
+    demographics_work_status: PropTypes.string,
+    demographics_current_work_sector: PropTypes.string,
+    demographics_future_work_sector: PropTypes.string,
+  }).isRequired,
+  saveSettings: PropTypes.func.isRequired,
+  updateDraft: PropTypes.func.isRequired
+};
+
+export default connect(demographicsSectionSelector, {
+  saveSettings,
```
oliviaruizknott, line 261: Note to self: don’t need `saveSettings`

src/account-settings/demographics/data/service.js
```diff
@@ -0,0 +1,124 @@
...
+ */
+export async function patchDemographics(userId, commitValues) {
+  const requestUrl = `http://localhost:18360/demographics/api/v1/demographics/${userId}/`;
+  const convertedCommitValues = convertData(commitValues, TO);
+  let data;
+
+  ({ data } = await getAuthenticatedHttpClient()
+    .patch(requestUrl, convertedCommitValues)
+    .catch((error) => {
+      // TODO: If the patch fails, try a POST first.
```
oliviaruizknott, line 74: This is one of the glaring holes. `PATCH` only works if the user exists; if not, do a `POST` first. (The first time anyone lands here they won’t exist in demographics yet (probably).

src/account-settings/data/sagas.js
```diff
@@ -106,19 +106,23 @@ export function* handleSaveSettings(action) {
...
       const { formId, commitValues } = settingsArray[i];
       yield put(saveSettingsBegin());
       const commitData = { [formId]: commitValues };
       const savedSettings = yield call(patchSettings, username, commitData, userId);
       yield put(saveSettingsSuccess(savedSettings, commitData));
     }
-    yield put(saveMultipleSettingsSuccess(settings));
+    yield put(saveMultipleSettingsSuccess(action));
+    if (formId) {
+      yield delay(1000);
```
justinhynes, line 123: The ``delay`` call caught my eye. Is this needed/necessary?

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -0,0 +1,264 @@
...
+  };
+
+  render() {
+    const editableFieldProps = {
+      onChange: this.handleEditableFieldChange,
+      onSubmit: this.handleSubmit,
+    };
+
+    const {
+      yearOfBirthOptions,
```
justinhynes, line 105: Can this be removed? Is this leftover from when we were still going to add a second "year of birth" question/field?

src/account-settings/demographics/data/service.js
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * get all data related to the demographics.
+ * @param {Number} userId users are identified in the api by LMS id
+ */
+export async function getDemographics(userId) {
+  let data = {};
+  try {
+    ({ data } = await getAuthenticatedHttpClient()
+      // TODO: Fix localhost hardcoding
+      .get(`http://localhost:18360/demographics/api/v1/demographics/${userId}`));
```
justinhynes: Just leaving this here while I am thinking of it. We will need to have a follow-up PR in the ``edx-internal`` repo as well that sets the "real" config values of whatever-parameter-name we decide on after for the Stage and Prod environments.

src/account-settings/demographics/DemographicsSection.jsx
oliviaruizknott: Eeep! Yes! Good catch.

src/account-settings/data/sagas.js
Tj-Tracy: Its following the pattern used in other places, its so that a person can register the fact the data was saved before it closes.

src/account-settings/EditableField.jsx
```diff
@@ -66,15 +67,20 @@ function EditableField(props) {
...
   const renderValue = (rawValue) => {
     if (!rawValue) return renderEmptyLabel();
+    let value = rawValue;
 
     if (options) {
       // Use == instead of === to prevent issues when HTML casts numbers as strings
       // eslint-disable-next-line eqeqeq
       const selectedOption = options.find(option => option.value == rawValue);
-      if (selectedOption) return selectedOption.label;
+      if (selectedOption) value = selectedOption.label;
```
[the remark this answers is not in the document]
oliviaruizknott: This is all passing the linter now (and you’ll notice it was a one-line statement before), but I’m very happy to drop it to two lines. I also find that more readable.

src/account-settings/EditableField.jsx
oliviaruizknott: This was thanks to you! ✨

src/account-settings/JumpNav.jsx
```diff
@@ -31,6 +32,11 @@ function JumpNav({ intl }) {
             {intl.formatMessage(messages['account.settings.section.profile.information'])}
           </NavHashLink>
         </li>
+        <li>
+          <NavHashLink to="#demographics-information">
+            {intl.formatMessage(messages['account.settings.section.demographics.information'])}
```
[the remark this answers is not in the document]
oliviaruizknott: Uh ... I’m not seeing what you’re talking about. Is it possible your window width might be too narrow? (Unless you’re saying I need to _add_ whitespace somewhere?)

src/account-settings/data/sagas.js
oliviaruizknott: ☝️ Yes @justinhynes, what TJ said. I just made this match the existing "closeForm" behavior.

src/account-settings/demographics/Checkboxes.jsx
```diff
@@ -0,0 +1,78 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { CheckBox } from '@edx/paragon';
+import { DECLINED } from '../data/constants';
+
+export const Checkboxes = (props) => {
```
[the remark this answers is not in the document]
oliviaruizknott: If I’m understanding your question, "Prefer not to respond" should not be able to be checked at the same time as any other option. Checking it should uncheck anything else; checking anything else when it’s checked should uncheck it. This is handled in the `handleToggle` function below. 

See slight demonstration at the ~30 second mark in this video: https://www.loom.com/share/ae6fdc6442434686a4511c03a5d83260

(Is that what you’re asking about here?)

src/account-settings/demographics/DemographicsSection.jsx
oliviaruizknott: Short answer: no, that doesn’t happen! 🌈 Opening any other field 1) closes the current field and 2) wipes out the drafts.

A demonstration. You can watch the `drafts` attribute in the Redux tools if you’re familiar with that interface, or I’m also console.logging the value of `drafts` for demonstration here.
https://www.loom.com/share/36a959c750d54434ae7d23fea67e0ee3

Now, this still may be ill-advised for other reasons. If there is work done to change the behavior of fields in the other sections, this may be silently affected if other developers don’t know about this? Given the recurring conversations we seem to be having about the architecture of the account page, I’m hopeful that any reworking would be an overhaul, and/or that we’d have some window into that.

If you have a better way to accomplish this, I’d be happy to explore that!

src/account-settings/demographics/DemographicsSection.jsx
oliviaruizknott: Oh niiiiiiice.

src/account-settings/demographics/DemographicsSection.jsx
Tj-Tracy: Thanks for clarifying, I think this will work for now. 👍

PR #245, opened 2020-06-18
IDV UI improvements

bseverino:
@edx/masters-devs-cosmonauts This PR is ready for review

See [MST-243](https://openedx.atlassian.net/browse/MST-243)

Completed:
- Fix padding issues on "next" button
- Expand/collapse help text on photo panels to avoid overwhelming the user
- Made name change and submit button functional

### DIFF
--- a/src/id-verification/data/service.js
+++ b/src/id-verification/data/service.js
@@ -45,8 +45,6 @@ export async function submitIdVerification(verificationData) {
     facePhotoFile: 'face_image',
     idPhotoFile: 'photo_id_image',
     idPhotoName: 'full_name',
-    courseRunKey: 'course_id',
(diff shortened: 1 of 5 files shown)


src/id-verification/data/service.js
```diff
@@ -45,7 +45,7 @@ export async function submitIdVerification(verificationData) {
     facePhotoFile: 'face_image',
     idPhotoFile: 'photo_id_image',
     idPhotoName: 'full_name',
-    courseRunKey: 'course_id',
+    // courseRunKey: 'course_id',
```
schenedx, line 48: Instead of comment this out, why not just remove?

bseverino: It should be implemented later, but I can log it in a ticket instead

PR #248, opened 2020-06-25
Hide demographics sidenav link when demographics is off

Tj-Tracy:
Turn of the side navigation when demographics is not turned on.

### DIFF
--- a/src/account-settings/JumpNav.jsx
+++ b/src/account-settings/JumpNav.jsx
@@ -4,6 +4,7 @@ import { NavHashLink } from 'react-router-hash-link';
 import Scrollspy from 'react-scrollspy';
 
 import messages from './AccountSettingsPage.messages';
+import { getConfig } from '@edx/frontend-platform';
 
 
 function JumpNav({ intl }) {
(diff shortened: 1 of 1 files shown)


src/account-settings/JumpNav.jsx
```diff
@@ -32,11 +33,13 @@ function JumpNav({ intl }) {
             {intl.formatMessage(messages['account.settings.section.profile.information'])}
           </NavHashLink>
         </li>
-        <li>
-          <NavHashLink to="#demographics-information">
-            {intl.formatMessage(messages['account.settings.section.demographics.information'])}
-          </NavHashLink>
-        </li>
+        {getConfig().ENABLE_DEMOGRAPHICS_COLLECTION &&
```
crice100, line 36: Is the trailing && just in case we add another check?

Tj-Tracy: Its a shortcut for conditional rendering. Its an abuse of `and` because a component rendering is always "truthy". This is saying "hey, if this conditional is true, render this". a little neater than `conditinal ? <component/> : null`

PR #251, opened 2020-06-30
MICROBA-309 | Error handling, tests, misc. updates for Demographics

justinhynes:
[MICROBA-309]

- Fix defect in ethnicityFieldDisplay function. Check to make sure the prop exists before trying to access.
- Add error handling for the scenario that the Demographics service isn't available/isn't responding to API calls. Show alert on PATCH calls to inform user that service is temporarily unavailable
- Add tests for DemographicsSection component

### DIFF
--- a/src/account-settings/demographics/DemographicsSection.jsx
+++ b/src/account-settings/demographics/DemographicsSection.jsx
@@ -1,31 +1,46 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
-import { Input } from '@edx/paragon';
-import memoize from 'memoize-one';
-
-import { demographicsSectionSelector } from '../data/selectors';
-import { saveMultipleSettings, updateDraft } from '../data/actions';
-import EditableField from '../EditableField';
-import Checkboxes from './Checkboxes';
-import messages from './DemographicsSection.messages';
 import {
-  SELF_DESCRIBE,
-  DEMOGRAPHICS_GENDER_OPTIONS,
+  DECLINED,
+  DEMOGRAPHICS_EDUCATION_LEVEL_OPTIONS,
   DEMOGRAPHICS_ETHNICITY_OPTIONS,
+  DEMOGRAPHICS_GENDER_OPTIONS,
   DEMOGRAPHICS_INCOME_OPTIONS,
   DEMOGRAPHICS_MILITARY_HISTORY_OPTIONS,
-  DEMOGRAPHICS_EDUCATION_LEVEL_OPTIONS,
-  OTHER,
-  DEMOGRAPHICS_WORK_STATUS_OPTIONS,
   DEMOGRAPHICS_WORK_SECTOR_OPTIONS,
-  DECLINED,
+  DEMOGRAPHICS_WORK_STATUS_OPTIONS,
+  OTHER,
+  SELF_DESCRIBE,
 } from '../data/constants';
+import {
+  FormattedMessage,
+  injectIntl,
+  intlShape,
+} from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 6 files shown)


src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -1,31 +1,54 @@
-import React from 'react';
```
justinhynes, line 1: I attempted to use an import auto-sorter through VS Code at one point and it caused more headaches than it was worth. So sorry if it looks like I was just randomly rearranging imports =P

src/account-settings/demographics/data/service.js
```diff
@@ -2,6 +2,23 @@ import { getAuthenticatedHttpClient } from '@edx/frontend-platform/auth';
 import { getConfig } from '@edx/frontend-platform';
 import { convertData, TO, FROM } from './utils';
 
+/**
+ *  If there was an error making the PATCH call then we create an apiError object containing a
```
justinhynes, line 6: I just realized this comment needs to be updated. This will get called on an error for a `POST` or a `PATCH`. We will call this method in both places.

src/account-settings/demographics/data/service.js
```diff
@@ -33,13 +50,33 @@ export async function getDemographics(userId) {
...
-
-    data = convertData(data, FROM);
   } catch (error) {
-    data = await postDemographics(userId);
+    const apiError = Object.create(error);
+    // if the API called resulted in this user receiving a 404 then follow up with a POST call to
+    // try and create the demographics entity on the backend
+    if (apiError.customAttributes.httpErrorStatus) {
+      if (apiError.customAttributes.httpErrorStatus === 404) {
+        data = await postDemographics(userId);
```
justinhynes, line 59: I didn't adjust this much. I am still only calling `postDemographics()` to create the entity in the back-end when receiving a `404` when doing the `GET` call on Accounts page load.

src/account-settings/demographics/data/service.js
```diff
@@ -33,13 +50,33 @@ export async function getDemographics(userId) {
...
+      }
+    } else {
+      data = {
```
justinhynes, line 62: Trying to pass back errors in a `fieldError` in the GET call would have required changes to the `fetchSettings` action/saga that had really awful side effects.

So, if there are any other errors, I just pass back a default/blank object. This will appear like the user has never answered any of the Demographics questions before. There was prior art for this approach.

src/account-settings/demographics/data/service.js
```diff
@@ -33,13 +50,33 @@ export async function getDemographics(userId) {
...
+        work_status_description: null,
+        current_work_sector: null,
+        future_work_sector: null,
+        user_ethnicity: [],
+      };
+    }
   }
 
-  return data;
+  return convertData(data, FROM);
```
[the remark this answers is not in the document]
justinhynes: I will take a second look. I reached out on Slack to confirm my understanding of the scenario we are trying to test. My initial run through did not encounter any problems/errors.

src/account-settings/demographics/data/service.js
```diff
@@ -33,13 +50,33 @@ export async function getDemographics(userId) {
...
+        gender_description: null,
+        income: null,
+        learner_education_level: null,
+        parent_education_level: null,
+        military_history: null,
+        work_status: null,
+        work_status_description: null,
+        current_work_sector: null,
+        future_work_sector: null,
+        user_ethnicity: [],
```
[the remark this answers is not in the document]
justinhynes: Yeah, I know what you mean. However, there is the prior art I had mentioned. This appears to be an accepted pattern in other places when we cannot get the data we are requesting.

src/account-settings/demographics/data/service.js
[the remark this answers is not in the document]
justinhynes: OK. Thanks for pointing this out! I realized that there was a chance to double convert data in a certain scenario. I reverted the change.

PR #261, opened 2020-07-14
MICROBA-474 | Gate Demographics questions and link via LMS API call

justinhynes:
[MICROBA-474]
- Add logic to show (or hide) the "Optional Information" (Demographics questions) section on the Account settings page via call to new LMS API
- Remove unused import from DemographicsSection.test.jsx

### DIFF
--- a/src/account-settings/data/service.js
+++ b/src/account-settings/data/service.js
@@ -152,8 +152,30 @@ export async function getProfileDataManager(username, userRoles) {
 }
 
 /**
+ * A function to determine if the Demographics questions should be displayed to the user. For the
+ * MVP release of Demographics we are limiting the Demographics question visibility only to
+ * MicroBachelors learners.
+ */
+export async function shouldDisplayDemographicsQuestions() {
+  const requestUrl = `${getConfig().LMS_BASE_URL}/api/demographics/v1/demographics/status/`;
+  let data = {};
+
+  try {
+    ({ data } = await getAuthenticatedHttpClient().get(requestUrl));
+    if (data.display) {
+      return data.display;
+    }
+  } catch (error) {
+    // if there was an error then we just hide the section
+    return false;
+  }
+
+  return false;
+}
+
+/**
  * A single function to GET everything considered a setting.
- * Currently encapsulates Account, Preferences, Coaching, and ThirdPartyAuth
+ * Currently encapsulates Account, Preferences, Coaching, ThirdPartyAuth, and Demographics
  */
 export async function getSettings(username, userRoles, userId) {
   const results = await Promise.all([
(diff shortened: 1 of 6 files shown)


src/account-settings/demographics/test/DemographicsSection.test.jsx
```diff
@@ -1,5 +1,4 @@
 import * as auth from '@edx/frontend-platform/auth';
-import * as selectors from '../../data/selectors';
```
justinhynes, line 2: I was referring back to this file when I was writing some new tests for the JumpNav component. I noticed that this import was not being used so I cleaned it up while I was in here.

src/account-settings/data/service.js
```diff
@@ -151,6 +151,28 @@ export async function getProfileDataManager(username, userRoles) {
   return null;
 }
 
+/**
+ * A function to determine if the Demographics questions should be displayed to the user. For the
+ * MVP release of Demographics we are limiting the Demographics question visibility only to
+ * MicroBachelors learners.
+ */
+export async function shouldDisplayDemographicsQuestions() {
+  const requestUrl = `${getConfig().LMS_BASE_URL}/api/demographics/v1/demographics/status/`;
```
staubina, line 160: Can we just verify how many times this is called on a render of the page?

src/account-settings/data/service.js
```diff
@@ -163,6 +185,7 @@ export async function getSettings(username, userRoles, userId) {
     getProfileDataManager(username, userRoles),
     getTimeZones(),
     getConfig().COACHING_ENABLED && getCoachingPreferences(userId),
+    getConfig().ENABLE_DEMOGRAPHICS_COLLECTION && shouldDisplayDemographicsQuestions(),
```
staubina, line 188: Will this still call the API if `ENABLE_DEMOGRAPHICS_COLLECTION` is False?

src/account-settings/data/service.js
```diff
@@ -151,6 +151,28 @@ export async function getProfileDataManager(username, userRoles) {
...
+    // if there was an error then we just hide the section
+    return false;
+  }
+
+  return false;
+}
+
 /**
  * A single function to GET everything considered a setting.
  * Currently encapsulates Account, Preferences, Coaching, and ThirdPartyAuth
```
staubina, line 178: Nit: Update the comment please.

src/account-settings/data/service.js
justinhynes: The `GET` call is made once, but the API is called twice on page render. First, there is an `OPTIONS` call and then a second for a `GET`.

The `OPTIONS` call is something I've noticed happens automagically before when using the AuthenticatedHttpClient provided by the `frontend-platform` code this repo is using. This is consistent behavior for all other API calls on this page as well (2 requests each -- an `OPTIONS` and a `GET` request).

staubina: That is what I would expect.

src/account-settings/data/service.js
justinhynes: I do not see the API call being made locally when Demographics collection is disabled.

src/account-settings/data/service.js
justinhynes: Thanks! Updating now.

PR #263, opened 2020-07-17
MICROBA-483 | Remove gender description as needed when gender identity is updated

justinhynes:
[MICROBA-483]
- Check if gender is updated before making the PATCH call for an update. Set gender_description to null if option is _not_ self-describe

### DIFF
--- a/src/account-settings/demographics/data/service.js
+++ b/src/account-settings/demographics/data/service.js
@@ -89,8 +89,17 @@ export async function getDemographics(userId) {
 export async function patchDemographics(userId, commitValues) {
   const requestUrl = `${getConfig().DEMOGRAPHICS_BASE_URL}/demographics/api/v1/demographics/${userId}/`;
   const convertedCommitValues = convertData(commitValues, TO);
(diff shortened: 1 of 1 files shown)


src/account-settings/demographics/data/service.js
```diff
@@ -89,8 +89,17 @@ export async function getDemographics(userId) {
 export async function patchDemographics(userId, commitValues) {
   const requestUrl = `${getConfig().DEMOGRAPHICS_BASE_URL}/demographics/api/v1/demographics/${userId}/`;
   const convertedCommitValues = convertData(commitValues, TO);
-  let data = {};
 
+  // Before patching we check to see if the gender options are being updated. If so, then we should
+  // try to make sure we have cleaned up any old data in case the user previously provided a
+  // self-description for their gender identity.
+  if ('gender' in convertedCommitValues) {
+    if (convertedCommitValues.gender !== 'self-describe') {
```
staubina, line 97: Can`'self-describe'` and `'gender'` be Constants?

PR #264, opened 2020-07-17
ID Verification: Hide image upload component

bseverino:
[MST-326](https://openedx.atlassian.net/browse/MST-326)

Temporary removal of image upload component from IDV flow. This component may be reimplemented at a later date with modifications.

- Camera access is required to successfully complete the process.
- If the user blocks camera access or camera access is otherwise unavailable, they will receive an error message and a button to return to the dashboard or course, depending on where they entered the IDV flow from.

----

Screen Shot 2020-07-17 at 1 23 36 PM


bseverino:
@Zacharis278 That's a good point, it'll be easier to switch back later. I'll do that instead.

PR #266, opened 2020-07-20
[MICROBA-484] Scroll to section based on URL

staubina:
Added in a check for a locationHash and loaded state to
scroll to the section in the URL so that the page would load to a Nav
Section after loading data.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -76,16 +76,15 @@ class AccountSettingsPage extends React.Component {
     });
   }
 
-  componentDidUpdate() {
-    if (this.props.loaded) {
+  componentDidUpdate(prevProps) {
+    if (prevProps.loading && !prevProps.loaded && this.props.loaded) {
       const locationHash = global.location.hash;
-
       // Check for the locationHash in the URL and then scroll to it if it is in the
       // NavLinks list
       if (typeof locationHash !== 'string')
         return;
       if (Object.keys(this.navLinkRefs).includes(locationHash) && this.navLinkRefs[locationHash].current) {
-          window.scrollTo(0, this.navLinkRefs[locationHash].current.offsetTop)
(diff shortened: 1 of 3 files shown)


src/account-settings/demographics/DemographicsSection.messages.jsx
```diff
@@ -231,11 +231,6 @@ const messages = defineMessages({
     defaultMessage: 'Some college, but no degree',
     description: 'The label for the some college, but no degree education level option.',
   },
-  'account.settings.field.demographics.education_level.options.some-college': {
```
staubina, line 234: NOTE: This was just cleanup when I noticed we had a duplicate entry.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -65,6 +76,20 @@ class AccountSettingsPage extends React.Component {
     });
   }
 
+  componentDidUpdate() {
+    if (this.props.loaded) {
```
davidjoy, line 80: Does this only execute once?  I'd be concerned it might force the user back down to the anchor tag whenever this component updates.  

Does it work to do this whole thing in componentDidMount instead, so you can ensure it only happens once at the beginning?

staubina: I will give this a try and also test out the update functionality.

staubina: Good catch! Changing now.

staubina: I need the check here, ComponentDidMount is hit before loading is done. So I need to use ComponentDidUpdate so that I see the transition from loading too loaded. 

I added in a new check for the state change. Let me know what you think @davidjoy.

src/account-settings/demographics/DemographicsSection.messages.jsx
oliviaruizknott: Thank you!

PR #280, opened 2020-07-24
Update Paragon and add loading state + spinner to IDV summary page

bseverino:
[MST-339: Add spinner to IDV submission](https://openedx.atlassian.net/browse/MST-339)

@edx/masters-devs-cosmonauts 

Spinner added and button disabled while uploading images to avoid duplicate submissions. Paragon dependency is updated from 7.1.5 => 9.1.0 with this PR.

---

Screen Shot 2020-07-24 at 3 53 30 PM

### DIFF
--- a/src/id-verification/panels/SummaryPanel.jsx
+++ b/src/id-verification/panels/SummaryPanel.jsx
@@ -1,6 +1,6 @@
-import React, { useContext } from 'react';
+import React, { useState, useContext } from 'react';
 import { history } from '@edx/frontend-platform';
-import { Input, Button } from '@edx/paragon';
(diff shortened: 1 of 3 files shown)


src/id-verification/panels/SummaryPanel.jsx
```diff
@@ -22,9 +22,11 @@ function SummaryPanel(props) {
     idPhotoName,
   } = useContext(IdVerificationContext);
   const nameToBeUsed = idPhotoName || nameOnAccount || '';
+  const [isLoading, setIsLoading] = useState(false);
```
schenedx, line 25: #NIT @bseverino isLoading is not a specific name. I would use `isSubmitting`

PR #291, opened 2020-08-12
Fix extra call to LMS when PATCHing demographics

Tj-Tracy:
We were making an extra call to the LMS when making a PATCH to the
demographics service which we don't need to do.

The keys for demographics fields we not being removed from the
accountSettings object properly, which made them be seen as changes,
which triggered the call to the LMS. I've added a constant
DEMOGRAPHICS_FIELDS that keys out those fields, removing the call.

### DIFF
--- a/src/account-settings/data/service.js
+++ b/src/account-settings/data/service.js
@@ -9,6 +9,7 @@ import { handleRequestError, unpackFieldErrors } from './utils';
 import { getThirdPartyAuthProviders } from '../third-party-auth';
 import { getCoachingPreferences, patchCoachingPreferences } from '../coaching/data/service';
 import { getDemographics, patchDemographics } from '../demographics/data/service';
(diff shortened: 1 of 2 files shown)


src/account-settings/data/service.js
```diff
@@ -211,8 +212,9 @@ export async function patchSettings(username, commitValues, userId) {
   // user/v1/preferences where it does update. This is the one we use.
   const preferenceKeys = ['time_zone'];
   const coachingKeys = ['coaching'];
+  const demographicsKeys = DEMOGRAPHICS_FIELDS;
```
staubina, line 215: Quick question: Why do we need to move this into a constant variable rather than just use the value as is?

Tj-Tracy: Just in case anything in the future changes, or we need to use this elsewhere in account, I'd rather have a constant here rather than using raw strings.

PR #293, opened 2020-08-13
stop camera once user successfully submits photos to IDVerification s…

MichaelRoytman:
[MST-361](https://openedx.atlassian.net/browse/MST-361)

- Stop camera access once user successfully submits photos to IDVerification service.
  - Once a camera component is unmounted, stop the camera.
    - If we do not do this, at the time that the user submits their images, there are two [MediaStreamTracks](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack) active, one for each `CameraPhoto` instance, and I have no way to stop them. It will appear as though we are continuing to user the user's camera.
  - Once the user submits their photos, close the video stream that we grabbed when enabling their camera.

### DIFF
--- a/src/id-verification/tests/panels/SummaryPanel.test.jsx
+++ b/src/id-verification/tests/panels/SummaryPanel.test.jsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import { Router } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
-import { render, cleanup, act, screen, fireEvent } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
+import { render, cleanup, act, screen, fireEvent, waitFor } from '@testing-library/react';
 import '@edx/frontend-platform/analytics';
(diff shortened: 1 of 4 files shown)


src/id-verification/tests/panels/SummaryPanel.test.jsx
```diff
@@ -49,5 +50,6 @@ describe('SummaryPanel', () => {
     const button = await screen.findByTestId('submit-button');
     fireEvent.click(button);
     expect(submitIdVerification).toHaveBeenCalled();
+    await waitFor(() => expect(contextValue.stopUserMedia).toHaveBeenCalled())
```
MichaelRoytman, line 53: I'm a little torn on including this test, since it seems at odds with the testing strategy recommended by React Testing Library. I don't think there is a way that I could simulate a user using the page and checking whether their camera is on. This feels like an implementation detail, but it's also a requirement, so, perhaps including this check is justified.

PR #296, opened 2020-08-18
MICROBA-534 | Remove hardcoded Demographics options, pull choices from Demographics API

justinhynes:
[MICROBA-534]
- Pull Demographics field choices from Demographics API to support I18N/L10N
- Cleanup constants and duplicated strings from MFE wherever possible
- Small refactor of alert code


src/account-settings/demographics/data/service.js
```diff
@@ -120,3 +120,68 @@ export async function patchDemographics(userId, commitValues) {
...
+ * retrieve the options
+ */
+export async function getDemographicsOptions() {
+  const requestUrl = `${getConfig().DEMOGRAPHICS_BASE_URL}/demographics/api/v1/demographics/`;
+  let data = {};
+
+  try {
+    ({ data } = await getAuthenticatedHttpClient().options(requestUrl));
+  } catch (error) {
+    /* TODO: We should spend more time to come up with a better solution to error handling here.
```
justinhynes, line 134: I tried to articulate this in the [ticket](https://openedx.atlassian.net/browse/MICROBA-534) comments as well. 

The work I have done is functional, I just don't love the error handling solutions available this second. This approach keeps the entire page from crashing in the event that the Demographics IDA is not responding, which we hope is not all that often.

If we want to hold off on merging/reviewing this until after I am back from PTO where I can demonstrate the issue in a parking lot or separate meeting then I am happy with that as well.

src/account-settings/demographics/data/service.js
```diff
@@ -120,3 +120,22 @@ export async function patchDemographics(userId, commitValues) {
...
+ * retrieve the options for each field from the Demographics API
+ */
+export async function getDemographicsOptions() {
+  const requestUrl = `${getConfig().DEMOGRAPHICS_BASE_URL}/demographics/api/v1/demographics/`;
+  let data = {};
+
+  try {
+    ({ data } = await getAuthenticatedHttpClient().options(requestUrl));
+  } catch (error) {
+    // We are catching and suppressing errors here on purpose. If an error occurs during the
```
justinhynes, line 134: Just wanted to call attention to this comment. The decisions to hide or display fields in the DemographicsSection will be based on whether or not this data was successfully retrieved from this function call.

I'd rather catch and re-throw an error here that can be handled downstream but the changes to support this during Fetch/Get settings calls was non-trivial.

At the very least I wanted to document why we were catching and swallowing errors here.

justinhynes:
This PR introduced new behavior if we cannot communicate with the Demographics API to retrieve the field options. If the demographicsOptions prop is empty (meaning we can't extract any choices to display for the questions) we will just hide the fields completely and display an error in the `Optional Information` section:
image

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -34,70 +26,80 @@ import messages from './DemographicsSection.messages';
...
-    if(this.checkFormForDemographicsErrors()) {
-      this.alertRef.current.focus();
-    }
+  /**
+   * Utility method that helps determine if we were able to retrieve the available options for
+   * the Demographics questions. Returns true if the `demographicsOptions` prop is _not_ empty,
+   * otherwise false. This prop being empty is indicative of a failure communicating with the
+   * Demographics IDA's API.
+   */
+  retrievedDemographicsOptions() {
```
tuchfarber, line 37: super nit: I think phrasing "check"-type functions with `is`, `has`, `have`, etc is helpful to differentiate between the action and the check that the action happened. (TL;DR I read this as `retrieveDemographicsOptions` a bunch of times while reviewing. lol) 

E.g. this could be `hasRetrievedDemographicsOptions()` or something similar.

src/account-settings/demographics/DemographicsSection.jsx
```diff
@@ -34,70 +26,80 @@ import messages from './DemographicsSection.messages';
...
-      const ethnicities = this.props.formValues.demographics_user_ethnicity;
+      const ethnicities = this.props.formValues.demographics_user_ethnicity;  
       return ethnicities.map((e) => {
-        if (e == DECLINED) {
-          return this.props.intl.formatMessage(messages[`account.settings.field.demographics.options.declined`]);
+        for (var option in demographicsEthnicityOptions) {
+          if (e === demographicsEthnicityOptions[option].value) {
+            return demographicsEthnicityOptions[option].label;
+          }
         }
```
tuchfarber, line 102: nit: This is perfectly fine as is. If you want to use some built-ins versus filtering yourself, you can use `filter`:
```
var matchingOption = demographicsEthnicityOptions.filter(option => option.value === e)[0];
return matchingOption && matchingOption.label;
```
The second line just returns `undefined` if case nothing in the array matches the filter, similar to how it works today by not returning anything.

src/account-settings/demographics/DemographicsSection.jsx
justinhynes: I always struggle finding the right name for these kind of methods. I like the suggestion! I'll add this.

src/account-settings/demographics/DemographicsSection.jsx
justinhynes: Thanks! This is really helpful.

PR #298, opened 2020-08-19
Added instructions for camera permissions

alangsto:
## [MST-364](https://openedx.atlassian.net/browse/MST-364)

@edx/masters-devs-cosmonauts @wittjeff 

Added browser specific instructions for enabling camera access if it has been detected that camera access has been denied. 

For Chrome:
Screen Shot 2020-08-19 at 4 09 56 PM

For Safari:
Screen Shot 2020-08-19 at 4 15 06 PM

For Firefox:
Screen Shot 2020-08-19 at 4 16 20 PM

### DIFF
--- a/src/id-verification/panels/RequestCameraAccessPanel.jsx
+++ b/src/id-verification/panels/RequestCameraAccessPanel.jsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useContext } from 'react';
 import { Link } from 'react-router-dom';
+import Bowser from 'bowser';
 import { getConfig } from '@edx/frontend-platform';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { injectIntl, intlShape, FormattedMessage } from '@edx/frontend-platform/i18n';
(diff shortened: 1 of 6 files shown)


src/id-verification/panels/RequestCameraAccessPanel.jsx
```diff
@@ -16,6 +17,65 @@ function RequestCameraAccessPanel(props) {
   const panelSlug = 'request-camera-access';
   const nextPanelSlug = useNextPanelSlug(panelSlug);
   const { tryGetUserMedia, mediaAccess, userId } = useContext(IdVerificationContext);
+  const browser = Bowser.parse(window.navigator.userAgent).browser.name;
+
+  let browserDirections = <></>;
+  if (browser === 'Internet Explorer') {
```
schenedx, line 23: #NIT there should be a better way to structure this code, right? Maybe use another component?

alangsto: I wasn't sure if it should be it's own component because there would only be text in the component, so no functionality. I agree that it would make the code more readable, so I incorporated your suggestion and now there is a separate component. Thanks for the feedback!

PR #300, opened 2020-08-24
Added option for ID image upload

alangsto:
## [MST-360](https://openedx.atlassian.net/browse/MST-360)

These changes allow learners to upload a photo of their ID rather than taking a photo with their device's camera. This option to upload is contained by a collapsible component on the summary page.

This is what a user will see on the ID image capture page (no changes):
Screen Shot 2020-08-24 at 3 19 53 PM

This is what a user will see once they reach the summary page:
Screen Shot 2020-08-24 at 3 20 24 PM

When they open the collapsible component:
Screen Shot 2020-08-24 at 3 29 15 PM

When they upload an image:
Screen Shot 2020-08-24 at 3 29 25 PM

### DIFF
--- a/src/id-verification/CameraHelpWithUpload.jsx
+++ b/src/id-verification/CameraHelpWithUpload.jsx
@@ -1,0 +1,53 @@
+import React, { useState, useContext } from 'react';
+import PropTypes from 'prop-types';
+import { Collapsible } from '@edx/paragon';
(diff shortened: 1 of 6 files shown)


schenedx:
The code here looks good to me. However, I am wondering if there are ways to introduce "friction" to learners not accessing upload so easily?

PR #302, opened 2020-08-28
added object tracking

alangsto:
## [MST-385](https://openedx.atlassian.net/browse/MST-385)

Added object tracking to IDV camera component. If this feature is enabled, users will see a blue or red box around their face. 

What users will see on the Portrait camera page:
Screen Shot 2020-08-31 at 9 30 40 AM

A spinner will appear while face detection is loading:
Screen Shot 2020-08-31 at 9 31 05 AM

What users will see on the ID camera page:
Screen Shot 2020-08-31 at 9 31 24 AM

Example of what the boxes look like:
ezgif com-video-to-gif (3)


src/id-verification/Camera.jsx
```diff
@@ -26,6 +31,87 @@ class Camera extends React.Component {
     this.cameraPhoto.stopCamera();
   }
 
+  startDetection() {
+    this.setState({ shouldDetect: !this.state.shouldDetect });
+    while (!this.state.hasLoaded) {
```
alangsto, line 36: I don't like this loop, but I wasn't sure of another way to implement this. The video has to have the data loaded in order call detectFromVideoFrame, otherwise it will throw an error. I basically want to wait to call detectFromVideoFrame until the video has loaded.

src/id-verification/Camera.jsx
```diff
@@ -26,6 +31,87 @@ class Camera extends React.Component {
...
+    model.estimateFaces(video).then((predictions) => {
+      if (this.state.shouldDetect) {
+        this.showDetections(predictions);
+
+        requestAnimationFrame(() => {
+          this.detectFromVideoFrame(model, video);
+        });
+      }
+    }, (ignore) => {
+      // will catch error when next panel is reached, so we want to ignore
```
alangsto, line 52: I also don't really like having an empty catch, but this throws an error once the "next" button is clicked on the panel. I think that's happening because this function is still running for a short time after the "next" button is clicked.

src/id-verification/Camera.jsx
```diff
@@ -26,6 +31,87 @@ class Camera extends React.Component {
...
+        }
+
+        // Draw the bounding box.
+        ctx.strokeStyle = this.state.outlineColor;
+        ctx.lineWidth = 4;
+        ctx.strokeRect(start[0], start[1], size[0], size[1]);
+      });
+    } catch (error) {
+      // error that will be thrown after next button is clicked, we can ignore it
+      if (!(error instanceof TypeError)) {
```
alangsto, line 97: Similar issue as the one mentioned above.

src/id-verification/Camera.jsx
```diff
@@ -54,29 +146,41 @@ class Camera extends React.Component {
       ? 'do-transition camera-flash'
       : 'camera-flash';
     return (
-      <div className='camera-outer-wrapper shadow'>
-        <div className='camera-wrapper'>
+      <div className="camera-outer-wrapper shadow">
+        <CheckBox
```
alangsto, line 150: I have this enabled by a checkbox so that users can choose whether they want to use it or not. The detection takes a few seconds to load, so I think adding a spinner next to the label would be good.

src/id-verification/tests/Camera.test.jsx
```diff
@@ -42,4 +43,22 @@ describe('SubmittedPanel', () => {
     fireEvent.click(button);
     expect(defaultProps.onImageCapture).toHaveBeenCalled();
   });
+
+  it('adds canvas when detection is enabled', async () => {
```
alangsto, line 47: This is the only test I have right now, I'm working on adding more.

src/id-verification/Camera.jsx
```diff
@@ -26,6 +32,83 @@ class Camera extends React.Component {
...
+      .then((values) => {
+        this.detectFromVideoFrame(values[0], this.videoRef.current);
+      })
+      .catch((error) => {
+        console.error(error);
+      });
+  }
+
+  detectFromVideoFrame = (model, video) => {
+    model.estimateFaces(video).then((predictions) => {
```
alangsto, line 51: This line causes an error when you reach the next page, I think because this function is still running for a short period of time after the video.ref is gone. I had originally added try/catch logic to ignore this error, but I feel like there's a better way to handle this.

src/id-verification/Camera.jsx
schenedx: Can we have a break out condition? Like after 5 seconds if `this.state.hasLoaded` was not turn on to `true`, we exit and return false or something?

src/id-verification/Camera.jsx
```diff
@@ -26,6 +33,94 @@ class Camera extends React.Component {
...
+
+        requestAnimationFrame(() => {
+          this.detectFromVideoFrame(model, video);
+        });
+      }
+    });
+  };
+
+  showDetections = (predictions) => {
+    let ctx;
```
schenedx, line 73: I don't like the variable name `ctx`. Can we have this as something more sensible like `canvasContext`?

src/id-verification/Camera.jsx
```diff
@@ -26,6 +33,94 @@ class Camera extends React.Component {
...
+      const xAdjustment = 70;
+      const yAdjustment = 55;
+      const start = [prediction.topLeft[0] - xAdjustment, prediction.topLeft[1] - yAdjustment];
+      const end = [prediction.bottomRight[0] - xAdjustment, prediction.bottomRight[1] - yAdjustment];
+      const size = [end[0] - start[0], end[1] - start[1]];
+
+      const features = prediction.landmarks;
+      let isInPosition = true;
+      ctx.fillStyle = 'blue';
+      for (let j = 0; j < features.length; j++) {
```
schenedx, line 88: Can you add some comments on why we need this loop?

schenedx: I think we need this for each detected face on the image, right?

alangsto: I can add some comments but also explain it here! Each prediction is a face detected, and the loop goes through the array of landmarks associated with each prediction. This example is from the blazeface repo:

```
`predictions` is an array of objects describing each detected face, for example:

    [
      {
        topLeft: [232.28, 145.26],
        bottomRight: [449.75, 308.36],
        probability: [0.998],
        landmarks: [
          [295.13, 177.64], // right eye
          [382.32, 175.56], // left eye
          [341.18, 205.03], // nose
          [345.12, 250.61], // mouth
          [252.76, 211.37], // right ear
          [431.20, 204.93] // left ear
        ]
      }
    ]
```
Within the loop I am determining if each landmark (i.e. if the right eye, nose, left ear, etc.) are within the correct bounds. Hope this helps!

alangsto:
@schenedx I added code and comments to address your feedback, thank you!

PR #307, opened 2020-09-01
Added object tracking with coco ssd model

alangsto:
## [MST-385](https://openedx.atlassian.net/browse/MST-385)

Similar to this [PR](https://github.com/edx/frontend-app-account/pull/302), except that the coco ssd model is used for es5 compatibility reasons. 

I don't think this model works as well as blazeface because coco ssd can only detect if the input contains a person, not a face, but it should be fine.

What users will see:

ezgif com-video-to-gif (2)


alangsto:
Closing this PR because the blazeface model is now working

PR #308, opened 2020-09-01
Add forgot password page.

waheedahmed:
VAN-17

Screen Shot 2020-09-02 at 2 29 53 PM

Screen Shot 2020-09-02 at 2 29 27 PM

screencapture-localhost-1997-login-2020-09-02-16_44_02

### DIFF
--- a/src/logistration/forgot-password/messages.js
+++ b/src/logistration/forgot-password/messages.js
@@ -1,0 +1,36 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
+  'logisration.forgot.password.page.heading': {
+    id: 'logisration.forgot.password.page.heading',
+    defaultMessage: 'Password assistance',
+    description: 'The page heading for the forgot password page.',
+  },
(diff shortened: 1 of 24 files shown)


src/logistration/forgot-password/messages.js
```diff
@@ -0,0 +1,36 @@
+import { defineMessages } from '@edx/frontend-platform/i18n';
+
+const messages = defineMessages({
```
zainab-amir, line 3: Just for my understanding, when do we use defineMessages vs FormatMessage? Is defineMessage used when we have repetitive occurrence of a string?

waheedahmed: @zainab-amir These are two different ways to declare messages for i18n, the only reason I choose `defineMessages` over `FormattedMessage` to keep strings at one place and out of other code for better readability. And yes, it could also be used to handle multiple occurrences of strings.

adeelehsan (approved):
looks good 👍

zainab-amir (approved):

PR #317, opened 2020-09-03
Added logic for adjusting image resolution

alangsto:
## For [CR-2544](https://openedx.atlassian.net/browse/CR-2544)

@edx/masters-devs-cosmonauts 

There are a few learners that are still encountering file size errors when attempting to submit their ID verification. My understanding is that when access to their camera is requested with an ideal resolution and facing mode, the request for that resolution or facing mode is rejected and it falls back to the default resolution and facing mode.

Because of this, additional logic should be added to ensure that if the camera resolution will result in an image with too high of a resolution, there is additional processing done in order to decrease the resolution and therefore decrease the file size.

### DIFF
--- a/src/id-verification/Camera.jsx
+++ b/src/id-verification/Camera.jsx
@@ -129,14 +129,36 @@ class Camera extends React.Component {
     if (this.state.dataUri) {
       return this.reset();
     }
+
     const config = {
-      sizeFactor: 1,
+      sizeFactor: this.getSizeFactor(),
     };
 
     this.playShutterClick();
     const dataUri = this.cameraPhoto.getDataUri(config);
     this.setState({ dataUri });
     this.props.onImageCapture(dataUri);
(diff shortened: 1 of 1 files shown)


src/id-verification/Camera.jsx
```diff
@@ -129,8 +129,26 @@ class Camera extends React.Component {
     if (this.state.dataUri) {
       return this.reset();
     }
+
+    let sizeFactor = 1;
+    const settings = this.cameraPhoto.getCameraSettings();
```
schenedx, line 134: Can you wrap this into its own function for easier reading?

schenedx (commented):
Only 1 small comment. Let me test this out

schenedx (approved):
Tested on my local machine using a 1080P camera and the flow works.
Let's try this change in prod and see.

PR #318, opened 2020-09-04
add playsInline attribute to video element to ensure it does not pop out into full screen video on iOS Safari

MichaelRoytman:
This adds the `playsInline` attribute to the video element displayed by the Camera. This prevents the video element from "popping out" into fullscreen mode on iOS Safari. This was tested with iOS 13.1.3.

Old Behavior
idv-video-old-behavior

Note that the image that is displayed when the video player is closed is the only image that can be taken with the camera, no matter where the camera is pointing.

New Behavior
idv-video-new-behavior


alangsto (approved):
So glad you found a fix! LGTM

schenedx (approved):
👍

PR #322, opened 2020-09-10
use 'environment' facing mode when using the camera in the ID photo c…

MichaelRoytman:
use 'environment' facing mode when using the camera in the ID photo context; this will open the rear facing camera on mobile

[facingMode documentation](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/facingMode)

Old Experience

idv-camera-direction-old

New Experience

idv-camera-direction-new

### DIFF
--- a/src/id-verification/Camera.jsx
+++ b/src/id-verification/Camera.jsx
@@ -25,7 +25,10 @@ class Camera extends React.Component {
 
   componentDidMount() {
     this.cameraPhoto = new CameraPhoto(this.videoRef.current);
-    this.cameraPhoto.startCamera(FACING_MODES.USER, { width: 640, height: 480 });
+    this.cameraPhoto.startCamera(
+      this.props.isPortrait ? FACING_MODES.USER : FACING_MODES.ENVIRONMENT, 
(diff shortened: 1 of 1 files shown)


Zacharis278 (approved):
👍

schenedx (approved):
👍

src/id-verification/Camera.jsx
```diff
@@ -238,7 +241,11 @@ class Camera extends React.Component {
 Camera.propTypes = {
   intl: intlShape.isRequired,
   onImageCapture: PropTypes.func.isRequired,
-  isPortrait: PropTypes.bool.isRequired,
+  isPortrait: PropTypes.bool,
```
alangsto, line 244: I was wondering why you changed this? Just trying to understand the design choice and if there is a pattern I should be following in the future!

alangsto (approved):
LGTM

src/id-verification/Camera.jsx
MichaelRoytman: That was actually an accident! Thank you for asking. I'll change that now.

PR #323, opened 2020-09-10
Added error messaging for image upload and idv submission

alangsto:
@edx/masters-devs-cosmonauts 

There was no file size limit for uploading an ID image. Error messaging has been added to alert users if they have attempted to upload a file that is larger than 10MB. This has been tested manually.

Screen Shot 2020-09-10 at 12 31 02 PM

There was also no feedback if there was an IDV submission error, so users were stuck on a screen with a loading spinner. I have added error messaging to alert a user if there was a submission error. I added a test for this.

Screen Shot 2020-09-10 at 1 16 23 PM

### DIFF
--- a/src/id-verification/ImageFileUpload.jsx
+++ b/src/id-verification/ImageFileUpload.jsx
@@ -1,28 +1,52 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useState } from 'react';
+import { intlShape } from '@edx/frontend-platform/i18n';
 import PropTypes from 'prop-types';
+import { Alert } from '@edx/paragon';
+import messages from './IdVerification.messages';
 
-export default function ImageFileUpload({ onFileChange }) {
+
+export default function ImageFileUpload({ onFileChange, intl }) {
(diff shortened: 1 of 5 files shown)


src/id-verification/ImageFileUpload.jsx
```diff
@@ -1,28 +1,51 @@
...
   const handleChange = useCallback((e) => {
     if (e.target.files.length === 0) {
       return;
     }
 
     const fileObject = e.target.files[0];
-    const fileReader = new FileReader();
-    fileReader.addEventListener('load', () => onFileChange(fileReader.result));
-    fileReader.readAsDataURL(fileObject);
+    if (fileObject.size < 10000000) {
```
schenedx, line 17: #NIT Can you move this into a const variable into the class?

schenedx (approved):
Just a NIT comment. Thank you for jumping on this issue!

src/id-verification/ImageFileUpload.jsx
alangsto: Yes, will update that!

PR #327, opened 2020-09-14
Code refactoring.

waheedahmed:
VAN-59


src/logistration/RegistrationPage.jsx
```diff
@@ -31,26 +31,22 @@ class RegistrationPage extends React.Component {
...
-      email: this.state.email,
-      username: this.state.username,
-      password: this.state.password,
-      name: this.state.name,
-      honor_code: true,
-      country: this.state.country,
-    };
-
-    this.props.registerNewUser(payload);
+    if (this.state.formValid) {
```
zainab-amir, line 36: what happens if the form is not valid?

waheedahmed: @zainab-amir Right now it does nothing but we can show the errors on the unfilled fields.

zainab-amir: okay

zainab-amir (approved):

zainab-amir (approved):
LGTM!

PR #328, opened 2020-09-15
redirection after login

adeelehsan:
VAN-8

### DIFF
--- a/src/logistration/LoginPage.jsx
+++ b/src/logistration/LoginPage.jsx
@@ -1,13 +1,24 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { Button, Input, ValidationFormGroup } from '@edx/paragon';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faFacebookF, faGoogle, faMicrosoft } from '@fortawesome/free-brands-svg-icons';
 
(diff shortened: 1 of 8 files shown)


waheedahmed (changes requested):

src/logistration/LoginPage.jsx
```diff
@@ -1,14 +1,25 @@
...
 import { faFacebookF, faGoogle, faMicrosoft } from '@fortawesome/free-brands-svg-icons';
 
 import { loginRequest } from './data/actions';
-import { forgotPasswordSelector } from './data/selectors';
+import { forgotPasswordSelector, loginRequestSelector } from './data/selectors';
 import ConfirmationAlert from './ConfirmationAlert';
 
+
+const LoginRedirect = (props) => {
+  const { success, redirectUrl } = props;
```
waheedahmed, line 15: I believe these will be available in `props. loginResult` and not directly from `props`?

adeelehsan: I am passing the props https://github.com/edx/frontend-app-account/pull/328/files#diff-25ed5a93ad76b3f62d8dbf031455130bR89

src/logistration/LoginPage.jsx
```diff
@@ -1,14 +1,25 @@
...
+  if (success) {
+    window.location.href = redirectUrl;
```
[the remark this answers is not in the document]
adeelehsan: Redirect is used to for the internal redirection.

PR #331, opened 2020-09-16
redirect after login

adeelehsan:
Redirect the learner after the login based on the argument in the url. 
Url can have two param's next and course id.

- If there's no param in url then learner would be redirected to dashboard.
- In case of course id it would be redirected to that course.
- In case of next learner would be redirected to the mentioned url in next. 

[VAN-4](https://openedx.atlassian.net/browse/VAN-4)

### DIFF
--- a/src/logistration/LoginPage.jsx
+++ b/src/logistration/LoginPage.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
+import { useLocation } from 'react-router-dom';
 import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
 import { FormattedMessage } from '@edx/frontend-platform/i18n';
 import { Button, Input, ValidationFormGroup } from '@edx/paragon';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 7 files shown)


src/logistration/LoginPage.jsx
```diff
@@ -1,14 +1,25 @@
...
 import ConfirmationAlert from './ConfirmationAlert';
 
+
+const LoginRedirect = (props) => {
+  const { success, redirectUrl } = props;
+  if (success) {
+    window.location.href = redirectUrl;
+    return <></>;
+  }
+  return null;
```
waheedahmed, line 20: We can replace this return with `return <></>;` and remove the one inside if clause.

waheedahmed (changes requested):
Basic LoginPage tests and snapshots are already added into `hack/log2` branch, you just need to add your test case into it. Please rebase with latest `hack/log2`

src/logistration/LoginPage.jsx
```diff
@@ -105,7 +115,8 @@ class LoginPage extends React.Component {
   render() {
     return (
       <React.Fragment>
-        <div className="d-flex justify-content-center logistration-container">
+        <LoginRedirect success={this.props.loginResult.success} redirectUrl={this.props.loginResult.redirectUrl} />
+        <div className="d-flex justify-content-center registration-container">
```
waheedahmed, line 119: Change the class name back to `registration-container`.

src/logistration/tests/__snapshots__/LoginPage.test.jsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`LoginPage should match default section snapshot 1`] = `
 <div
-  className="d-flex justify-content-center logistration-container"
+  className="d-flex justify-content-center registration-container"
```
waheedahmed, line 5: Same here `registration-container`.

waheedahmed (approved):
Couple of minor nits, otherwise 👍

PR #332, opened 2020-09-16
Add realtime feedback for screenreaders during face detection

alangsto:
## [MST-386](https://openedx.atlassian.net/browse/MST-386)

Realtime feedback for screen readers has been added to the face detection feature to make IDV more accessible for visually impaired users. 

Position related feedback will only be given to users if exactly one face is detected. If one face is detected a user will be told if they are in a good or bad position. If they are in a bad position, they will also be told what area they are in (i.e. "Incorrect position. Bottom left."). 

Link to video example: https://imgur.com/a/UrsjXFn

### DIFF
--- a/src/id-verification/Camera.jsx
+++ b/src/id-verification/Camera.jsx
@@ -20,14 +20,16 @@ class Camera extends React.Component {
       videoHasLoaded: false,
       shouldDetect: false,
       isFinishedLoadingDetection: true,
+      shouldGiveFeedback: true,
+      feedback: '',
     };
   }
 
   componentDidMount() {
     this.cameraPhoto = new CameraPhoto(this.videoRef.current);
     this.cameraPhoto.startCamera(
-      this.props.isPortrait ? FACING_MODES.USER : FACING_MODES.ENVIRONMENT, 
-      { width: 640, height: 480 }
+      this.props.isPortrait ? FACING_MODES.USER : FACING_MODES.ENVIRONMENT,
+      { width: 640, height: 480 },
     );
   }
 
@@ -97,11 +99,14 @@ class Camera extends React.Component {
         const x = features[j][0];
         const y = features[j][1];
 
+        let isInRange;
         if (this.props.isPortrait) {
-          isInPosition = isInPosition && this.isInRangeForPortrait(x, y);
+          isInRange = this.isInRangeForPortrait(x, y);
         } else {
-          isInPosition = isInPosition && this.isInRangeForID(x, y);
+          isInRange = this.isInRangeForID(x, y);
         }
+        // if it is not in range, give feedback depending on which feature is out of range
+        isInPosition = isInPosition && isInRange;
       }
 
       // draw a box depending on if all landmarks are in position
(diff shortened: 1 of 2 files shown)


src/id-verification/IdVerification.messages.js
```diff
@@ -96,6 +96,66 @@ const messages = defineMessages({
...
+    description: 'Text for screen reader when face is in a bad position.',
+  },
+  'id.verification.photo.feedback.center.left': {
+    id: 'id.verification.photo.feedback.center.left',
+    defaultMessage: 'Incorrect position. Center left.',
+    description: 'Text for screen reader when face is in a bad position.',
+  },
+  'id.verification.photo.feedback.center.center': {
+    id: 'id.verification.photo.feedback.center.center',
+    defaultMessage: 'Incorrect position. Center.',
```
wittjeff, line 136: What is this scenario?  Objective is too small? If so we should be more specific so they can infer what to do to fix it.

alangsto: This would occur if someone were too close to the camera, I'll update the message!

src/id-verification/Camera.jsx
```diff
@@ -97,23 +99,89 @@ class Camera extends React.Component {
...
+        let isInRange;
         if (this.props.isPortrait) {
-          isInPosition = isInPosition && this.isInRangeForPortrait(x, y);
+          isInRange = this.isInRangeForPortrait(x, y);
         } else {
-          isInPosition = isInPosition && this.isInRangeForID(x, y);
+          isInRange = this.isInRangeForID(x, y);
         }
+        // if it is not in range, give feedback depending on which feature is out of range
+        isInPosition = isInPosition && isInRange;
```
schenedx, line 109: Nice refactoring

src/id-verification/Camera.jsx
```diff
@@ -97,23 +99,89 @@ class Camera extends React.Component {
...
+        newFeedback = this.props.intl.formatMessage(messages['id.verification.photo.feedback.two.faces']);
+      } else {
+        newFeedback = this.props.intl.formatMessage(messages['id.verification.photo.feedback.no.faces']);
+      }
+      if (currentFeedback !== newFeedback) {
+        // only update status if it is different, so we don't overload the user with status updates
+        this.setState({ feedback: newFeedback });
+      }
+      this.setState({ shouldGiveFeedback: false });
+      setTimeout(() => {
```
schenedx, line 154: Could you explain why you need this?

src/id-verification/Camera.jsx
```diff
@@ -97,23 +99,89 @@ class Camera extends React.Component {
...
+
+  getGridPosition(coordinates) {
+    // Used to determine where a face is (i.e. top-left, center-right, bottom-center, etc.)
+
+    const x = coordinates[0];
+    const y = coordinates[1];
+
+    let messageBase = 'id.verification.photo.feedback';
+
+    if (y < 160) {
```
schenedx, line 168: The numbers should have variable labels like:
`FaceHeightLimit = 160`

schenedx (approved):
Only minor feedback. Good work.

src/id-verification/Camera.jsx
alangsto: I had a timeout set here because without it, anytime you move feedback will be given. This causes feedback to be interrupted anytime a user moves, so I thought it would be more clear to insure that only one instruction is given at a time, and that it is uninterrupted.

wittjeff: Cadence for audio feedback is an interesting usability question. When I talked to blind people about the real-time feedback they get with the iPhone's camera app, they seemed to be under the impression that the feedback is event-triggered, yet it doesn't do this predictable interrupting of itself. So I don't know how they handle it but it seems there is room for innovation like Ali's approach here.

PR #333, opened 2020-09-17
suppressed pii for hot jar

adeelehsan:
VAN-23

### DIFF
--- a/src/account-settings/EditableField.jsx
+++ b/src/account-settings/EditableField.jsx
@@ -146,11 +146,11 @@ function EditableField(props) {
           </form>
         ),
         default: (
-          <div className="form-group">
+          <div data-hj-suppress className="form-group">
             <div className="d-flex align-items-start">
-              <h6 aria-level="3">{label}</h6>
+              <h6 data-hj-allow aria-level="3">{label}</h6>
               {isEditable ? (
-                <Button onClick={handleEdit} className="ml-3 btn-link">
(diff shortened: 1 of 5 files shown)


waheedahmed (commented):
Looks good 👍,  just confirm the demographics part. Also, do we need to upgrade the header component version for this in remote config or here in this repo to suppress the username in header?

waheedahmed (approved):

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -465,7 +465,7 @@ class AccountSettingsPage extends React.Component {
 
         <div className="account-section" id="linked-accounts" ref={this.navLinkRefs['#linked-accounts']}>
           <h2 className="section-heading">{this.props.intl.formatMessage(messages['account.settings.section.linked.accounts'])}</h2>
-          <p>{this.props.intl.formatMessage(messages['account.settings.section.linked.accounts.description'])}</p>
+          <p data-hj-suppress>{this.props.intl.formatMessage(messages['account.settings.section.linked.accounts.description'])}</p>
```
waheedahmed, line 468: Why we are suppressing a help text message?

waheedahmed (changes requested):
There are couple of `EmailField` references unsuppressed and what about the dropdown fields in `EditableField` component, does HotJar cover those automatically or do we need to suppress those? And social media links are also unsuppressed.

waheedahmed (approved):

PR #334, opened 2020-09-17
Move blazeface package to correct version

alangsto:
## [MST-409](https://openedx.atlassian.net/browse/MST-409)

Blazeface was previously being installed from a forked repo in order for it to be es5 compatible. There was a recent PR merge in the tfjs/models repo that allows blazeface to now be es5 compatible.


alangsto:
Still not es5 compatible, will reopen when problem is addressed

PR #336, opened 2020-09-24
Add need help links on forgot password page.

waheedahmed:
Moved login page need help links into its own component and used that on both login and forgot password pages.

VAN-17


zainab-amir (approved):
Just one comment, should we move the message to it's own file `LoginHelpLinks.messages.js`?

waheedahmed:
@zainab-amir Yes we can, let me move them into it's own file.

zainab-amir (approved):

PR #337, opened 2020-10-01
Add tracking events for face detection

alangsto:
## [MST-454](https://openedx.atlassian.net/browse/MST-454)

Added tracking events for when users enable and disable face detection for both the ID and portrait photo pages.


schenedx (approved):
👍

PR #339, opened 2020-10-20
Increase camera resolution to 1080p for ID verification

bseverino:
### [MST-490](https://openedx.atlassian.net/browse/MST-490)

**Note:** Safari users are still limited to 480p as it doesn't handle the face detection feature correctly.


alangsto (approved):
LGTM

PR #349, opened 2020-11-10
fix: Update frontend-build and frontend-platform

davidjoy:
This is a local copy of https://github.com/edx/frontend-app-account/pull/340.  We think that PR wouldn't complete its travis checks because it's from a fork, for some crazy reason.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,23 +5,34 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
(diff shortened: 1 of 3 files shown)


davidjoy:
@morenol Let's see if this works.

nedbat (approved):

davidjoy:
Guess we don't need this!  Travis let #340 finish.  Merged that instead.

PR #351, opened 2020-11-14
Prevent IDV access from audit courses

bseverino:
### [MST-513](https://openedx.atlassian.net/browse/MST-513)

This PR introduces an API call to check whether the learner is enrolled in at least one verified course mode, and blocks access to IDV if they are not. The `ExistingRequest` component has been repurposed to `AccessBlocked` to allow for multiple instances where the learner is denied access to IDV (existing approved verification, unverified account, etc).

---
Screen Shot 2020-11-16 at 12 45 33 PM


src/id-verification/IdVerificationContext.jsx
```diff
@@ -16,6 +16,14 @@ const MEDIA_ACCESS = {
   GRANTED: 'granted',
 };
 
+const ERROR_REASONS = {
+  AUDIT: 'audit',
```
schenedx, line 20: I rather say the reason is `COURSE_ENROLLMENT`. This way, it's not just AUDIT track.

src/id-verification/IdVerificationContext.jsx
```diff
@@ -61,24 +71,46 @@ function IdVerificationContextProvider({ children }) {
     },
   };
 
-  // Call verification status endpoint to check whether we can verify.
   useEffect(() => {
```
schenedx, line 74: Having two `useEffect` call in the same file is necessary? Can we combine them into one?
If having two is necessary, can we have one to call `getEnrollment()` and the other to call `getExistingIDVerification()`? That makes this code block easier to read.

src/id-verification/tests/AccessBlocked.test.jsx
```diff
@@ -37,29 +39,29 @@ describe('ExistingRequest', () => {
     expect(text).toBeInTheDocument();
   });
 
-  it('renders correctly when status is approved', async () => {
-    defaultProps.status = 'approved';
+  it('renders correctly when learner is in audit mode', async () => {
```
schenedx, line 42: `renders correctly when learner is not in enrollment requiring IDV`

schenedx (approved):
👍 All inline comments NIT.

PR #355, opened 2020-12-04
Revert "Prevent IDV access from audit courses"

bseverino:
Reverts edx/frontend-app-account#351

Temporary revert while we investigate 400 errors


schenedx (approved):
👍

PR #358, opened 2020-12-07
Parse course_id query string correctly

bseverino:
The course_id was being parsed incorrectly in the IDV flow, preventing users from submitting their photos. Related to https://github.com/edx/edx-platform/pull/25792


schenedx (approved):
👍

PR #359, opened 2020-12-09
Remove use of course key temporarily

bseverino:
The course key is being parsed incorrectly, leading to 400 errors. This is a temporary fix while we investigate the solution.


schenedx (approved):
👍

PR #360, opened 2020-12-10
[MST-567] Allow decoding of query string in IDV flow

bseverino:
[MST-567](https://openedx.atlassian.net/browse/MST-567)

Related to https://github.com/edx/edx-platform/pull/25845 and https://github.com/edx/ecommerce/pull/3284

This addition decodes the `course_id` query string passed into the IDV flow, allowing special characters like `+`.

### DIFF
--- a/src/id-verification/IdVerificationPage.jsx
+++ b/src/id-verification/IdVerificationPage.jsx
@@ -30,8 +30,11 @@ function IdVerificationPage(props) {
   // Course run key is passed as a query string
   useEffect(() => {
     if (search) {
-      const parsed = qs.parse(search, { ignoreQueryPrefix: true });
-      if (parsed.hasOwnProperty('course_id')) {
+      const parsed = qs.parse(search, {
+        ignoreQueryPrefix: true,
+        interpretNumericEntities: true
+      });
+      if (parsed.hasOwnProperty('course_id') && parsed.course_id) {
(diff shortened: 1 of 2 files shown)


src/id-verification/IdVerificationPage.jsx
```diff
@@ -30,7 +30,11 @@ function IdVerificationPage(props) {
   // Course run key is passed as a query string
   useEffect(() => {
     if (search) {
-      const parsed = qs.parse(search, { ignoreQueryPrefix: true });
+      const parsed = qs.parse(search, {
+        ignoreQueryPrefix: true,
+        charset: 'iso-8859-1',
+        interpretNumericEntities: true
```
schenedx, line 36: Why do we need this flag?

schenedx (changes requested):
Is there a way to unit test this?
For example, after mounting this jsx, check in the sessionStorage?

bseverino:
@schenedx Thanks for reminding me, I was going to see if there was a way I could do that. I'll work on that now.

src/id-verification/IdVerificationPage.jsx
bseverino: This is needed to convert the encoded query string, IE `course-v1%3AedX%2BDemoX%2BDemo_Course` is decoded back into `course-v1:edX+DemoX+Demo_Course`

schenedx (approved):
👍

PR #363, opened 2020-12-14
[MST-567] Re-enable use of course key in IDV flow

bseverino:
[MST-567](https://openedx.atlassian.net/browse/MST-567)

If the learner accesses the IDV flow through a course, this redirects them back to that course after they are finished with the verification process, rather than the dashboard. This does **not** re-enable the ability to send the course key back while submitting photos, as this is unneeded in the current flow.


bseverino:
We decided to put this on hold for now.

PR #368, opened 2020-12-16
[MST-494] Clarify photo ID instructions with example image

bseverino:
[MST-494](https://openedx.atlassian.net/browse/MST-494)

This adds an example image of a photo ID card and updates the copy in order to clarify what photo ID means. I've also removed an outdated comment on the submitted panel.

---
Screen Shot 2020-12-16 at 1 09 12 PM

---
Screen Shot 2020-12-16 at 1 43 07 PM


src/id-verification/IdVerification.messages.js
```diff
@@ -33,7 +38,7 @@ const messages = defineMessages({
   },
   'id.verification.requirements.card.id.text': {
     id: 'id.verification.requirements.card.id.text',
-    defaultMessage: 'You need a valid ID that contains your full name and photo.',
+    defaultMessage: 'You need a valid identification card that contains your full name and photo.',
```
MichaelRoytman, line 41: I could be wrong here, but I vaguely recall that if a particular message has been translated for a given key, then modifying the `defaultMessage`  for that key doesn't result in the message being re-translated. I recall needing to specify a new key. This is my memory from an issue from three years ago, though, so I could be off base. Do you know anything about that?

bseverino: Not sure but it sounds likely - I'll see if there's anything in the docs

bseverino: Okay, based on the info here it should flag it to be translated again if the value changes but the key stays the same: https://docs.transifex.com/projects/updating-content

MichaelRoytman: Great. Thank you for checking!

MichaelRoytman (approved):

PR #411, opened 2021-03-09
[MST-655] Add Optimizely to be used with IDV experiments

bseverino:
[MST-655](https://openedx.atlassian.net/browse/MST-655)

Related backend PR: https://github.com/edx/edx-platform/pull/26925

### DIFF
--- a/src/id-verification/panels/ReviewRequirementsPanel.jsx
+++ b/src/id-verification/panels/ReviewRequirementsPanel.jsx
@@ -11,15 +11,29 @@ import messages from '../IdVerification.messages';
 import exampleCard from '../assets/example-card.png';
 
 function ReviewRequirementsPanel(props) {
-  const { userId } = useContext(IdVerificationContext);
+  const { userId, setOptimizelyExperimentName } = useContext(IdVerificationContext);
   const panelSlug = 'review-requirements';
   const nextPanelSlug = useNextPanelSlug(panelSlug);
+
+  const getExperiments = () => {
+    const {
+      experimentVariables: {
+        experimentName = '',
+      } = {},
+    } = window;
+
+    if (experimentName) {
+      setOptimizelyExperimentName(experimentName);
+    }
+  };
 
   useEffect(() => {
     sendTrackEvent('edx.id_verification.started', {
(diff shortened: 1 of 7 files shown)


src/id-verification/panels/ReviewRequirementsPanel.jsx
```diff
@@ -11,15 +11,29 @@ import messages from '../IdVerification.messages';
 import exampleCard from '../assets/example-card.png';
 
 function ReviewRequirementsPanel(props) {
-  const { userId } = useContext(IdVerificationContext);
+  const { userId, setOptimizelyExperimentName } = useContext(IdVerificationContext);
   const panelSlug = 'review-requirements';
   const nextPanelSlug = useNextPanelSlug(panelSlug);
 
+  const getExperiments = () => {
```
alangsto, line 18: Is there any way we can move this code to `src/id-verification/IdVerificationContextProvider.jsx`? I'm thinking it might be more useful to have that variable available in the context so it could be used in multiple panels.

bseverino: This will set it in context at the start of the flow, so it should persist (although I should probably add tests to ensure that's always the case, which I can do now)

alangsto: I think I was suggesting that the actual value be set in the Context, because a user won't reach the summary panel until the end of the flow. Based on my understanding, that would mean that if we wanted to use the `optimizelyExperimentName` in another panel that comes before the summary, we would need to move this code to that panel as well. That's why I suggested moving lines 18-29 to the context file. Let me know if I'm not understanding something correctly!

bseverino: Oh, right! This is the first panel, not the summary panel, so it will be updated in context as soon as the user enters the flow (if the experiment is on).

bseverino: That being said, it might make more sense for it to exist in the actual context provider, so I'll see if it still functions the same way.

alangsto (approved):

PR #413, opened 2021-03-12
[MST-704] Make name parameter optional in IDV submission

bseverino:
[MST-704](https://openedx.atlassian.net/browse/MST-704)

Even though `full_name` is an optional parameter that should only be submitted when the user is updating their account name, it is currently being sent with every IDV submission. This results in errors when the name is somehow blank, or the learner’s account information is controlled by a third party and can’t be updated through our API.

This update ignores the parameter if the value is `''` or `null`.

### DIFF
--- a/src/id-verification/tests/panels/SummaryPanel.test.jsx
+++ b/src/id-verification/tests/panels/SummaryPanel.test.jsx
@@ -91,6 +91,34 @@ describe('SummaryPanel', () => {
     await waitFor(() => expect(contextValue.stopUserMedia).toHaveBeenCalled());
   });
 
+  it('does not submit a name if name is blank', async () => {
+    contextValue.idPhotoName = '';
+    const verificationData = {
+      facePhotoFile: contextValue.facePhotoFile,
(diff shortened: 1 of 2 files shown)


src/id-verification/tests/panels/SummaryPanel.test.jsx
```diff
@@ -91,6 +91,27 @@ describe('SummaryPanel', () => {
...
+      optimizelyExperimentName: contextValue.optimizelyExperimentName,
+      courseRunKey: null,
+    };
+    await getPanel();
+    let button = await screen.findByTestId('submit-button');
+    fireEvent.click(button);
+    expect(dataService.submitIdVerification).toHaveBeenCalledWith(verificationData);
+    cleanup();
+
+    contextValue.idPhotoName = null;
```
schenedx, line 108: Please separate this into its own test case. Putting two cases into one is a bad practice because one case failure might masquerade the result of the other case.

bseverino:
Closing this as the changes have been merged, but the PR failed to close due to GitHub issues.

PR #416, opened 2021-03-19
MST-652: Add option to upload photos if user is in experiment

alangsto:
## [MST-652](https://openedx.atlassian.net/browse/MST-652)

These changes allow a user in an experiment to choose if they would like to upload photos instead of using their camera to submit IDV photos. I've introduced a new panel here which allows them to initially make that choice: 

Screen Shot 2021-03-19 at 9 04 29 AM

If they select that they would like to upload photos, they are immediately taken to this panel, bypassing the context panel that is usually shown before the photo capture pages: 

Screen Shot 2021-03-19 at 10 40 07 AM

As you can see on the above image, the user also has the option to switch between camera/upload mode throughout the IDV flow. The placement of the toggle button to switch between modes was suggested by Kevin on UX. If they choose to use their camera, they will see this panel: 

Screen Shot 2021-03-19 at 9 05 00 AM

I've also added additional routing to ensure that if a user tries to toggle to the camera mode, but hasn't enabled camera access yet, they will be routed to the panel that requests camera access. After camera access is enabled, they will be routed back to the portrait or ID photo capture pages.

### DIFF
--- a/src/id-verification/ImageFileUpload.jsx
+++ b/src/id-verification/ImageFileUpload.jsx
@@ -27,7 +27,7 @@ export default function ImageFileUpload({ onFileChange, intl }) {
     <>
       <input
         type="file"
-        accept="image/*"
+        accept=".png, .jpg, .jpeg"
         data-testid="fileUpload"
         onChange={handleChange}
       />
--- a/src/id-verification/panels/TakeIdPhotoPanel.jsx
+++ b/src/id-verification/panels/TakeIdPhotoPanel.jsx
@@ -9,24 +9,43 @@ import IdVerificationContext from '../IdVerificationContext';
 
 import messages from '../IdVerification.messages';
 import CameraHelp from '../CameraHelp';
+import ImagePreview from '../ImagePreview';
+import ImageFileUpload from '../ImageFileUpload';
+import CollapsibleImageHelp from '../CollapsibleImageHelp';
 
 function TakeIdPhotoPanel(props) {
   const panelSlug = 'take-id-photo';
   const nextPanelSlug = useNextPanelSlug(panelSlug);
-  const { setIdPhotoFile, idPhotoFile } = useContext(IdVerificationContext);
+  const {
+    setIdPhotoFile, idPhotoFile, optimizelyExperimentName, shouldUseCamera,
+  } = useContext(IdVerificationContext);
 
   return (
     <BasePanel
       name={panelSlug}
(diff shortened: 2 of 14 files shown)


src/id-verification/ImageFileUpload.jsx
```diff
@@ -27,7 +27,7 @@ export default function ImageFileUpload({ onFileChange, intl }) {
     <>
       <input
         type="file"
-        accept="image/*"
+        accept=".png, .jpg, .jpeg"
```
schenedx, line 30: Thank you!

src/id-verification/panels/TakeIdPhotoPanel.jsx
```diff
@@ -9,24 +9,43 @@ import IdVerificationContext from '../IdVerificationContext';
...
       name={panelSlug}
-      title={props.intl.formatMessage(messages['id.verification.id.photo.title.camera'])}
+      title={shouldUseCamera ? props.intl.formatMessage(messages['id.verification.id.photo.title.camera']) : props.intl.formatMessage(messages['id.verification.id.photo.title.upload'])}
     >
       <div>
-        <p>
-          {props.intl.formatMessage(messages['id.verification.id.photo.instructions.camera'])}
-        </p>
-        <Camera onImageCapture={setIdPhotoFile} isPortrait={false} />
+        {idPhotoFile && !shouldUseCamera && <ImagePreview src={idPhotoFile} alt={props.intl.formatMessage(messages['id.verification.id.photo.preview.alt'])} />}
```
schenedx, line 29: Can you help me understand this change? Is it when the user finished capturing or uploading their face photo, they will arrive here and have a choice to upload or capture their ID photo?

schenedx (approved):
Could you make sure, if the user is in the "control" group of the experiment, the learner sees the old experience, even with the new URLs? This is an edge case I can think of.

src/id-verification/panels/TakeIdPhotoPanel.jsx
alangsto: The change to the ID Photo Panel is to mirror the logic that is on the Portrait Photo Panel. This specific line is used to display the photo that the user has uploaded.

alangsto:
@schenedx I added some code to the routing utilities file to ensure that a user will be redirected if they try to access the new URL I created. Let me know if that addresses the edge case you were thinking of

PR #418, opened 2021-03-22
[MST-535] Prevent IDV name change if name is managed by a third party

bseverino:
[MST-535](https://openedx.atlassian.net/browse/MST-535)

Rework of https://github.com/edx/frontend-app-account/pull/414, fixed so that `getProfileDataManager` is returns a value rather than a promise.

### DIFF
--- a/src/id-verification/panels/GetNameIdPanel.jsx
+++ b/src/id-verification/panels/GetNameIdPanel.jsx
@@ -1,10 +1,11 @@
 import React, {
   useContext, useState, useEffect, useRef,
 } from 'react';
-import { Form } from '@edx/paragon';
+import { getConfig } from '@edx/frontend-platform';
+import { Hyperlink, Form } from '@edx/paragon';
 import { Link, useHistory } from 'react-router-dom';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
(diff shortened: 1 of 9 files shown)


src/id-verification/panels/GetNameIdPanel.jsx
```diff
@@ -45,14 +46,48 @@ function GetNameIdPanel(props) {
     }
   }, [nameMatches, blankName]);
 
-  const handleSubmit = (e) => {
+  function getNameValue() {
+    if (!nameMatches) {
+      // If we just check that idPhotoName exists, an empty string will cause
+      // the field to reset to nameOnAccountValue, so check that it is a string here.
+      if (typeof idPhotoName === 'string') {
```
schenedx, line 53: It just gives me pause to see this check. Shouldn't we just check if `idPhotoName == null`? That is more specific than a type check right? And easier to read.

schenedx (approved):
My only comment is a NIT. LGTM.

src/id-verification/panels/GetNameIdPanel.jsx
bseverino: Yeah, that makes more sense

PR #419, opened 2021-03-26
Remove option to upload if user is in experiment

alangsto:
## [MST-654](https://openedx.atlassian.net/browse/MST-654)
@edx/masters-devs-cosmonauts 

If a user is part of an A/B experiment for IDV, they will not be shown the option to upload an ID photo on the summary panel. 

The current experience looks like this: 
Screen Shot 2021-03-26 at 11 17 00 AM

The experiment experience will look like this:
Screen Shot 2021-03-26 at 11 16 23 AM


schenedx (approved):
👍

PR #422, opened 2021-03-30
[MST-718] Add extra validation to IDV image upload

bseverino:
[MST-718](https://openedx.atlassian.net/browse/MST-718)

As an extra measure to prevent the user from uploading unsupported file types such as `.pdf`, this adds JS validation on top of the `accept` attribute.

### DIFF
--- a/src/id-verification/IdVerification.messages.js
+++ b/src/id-verification/IdVerification.messages.js
@@ -413,7 +413,7 @@ const messages = defineMessages({
   },
   'id.verification.portrait.photo.instructions.upload': {
     id: 'id.verification.portrait.photo.instructions.upload',
-    defaultMessage: 'Please upload a portrait photo. Ensure your entire face fits inside the frame and is well-lit. (Supported formats: .jpg, .jpeg, .png)',
+    defaultMessage: 'Please upload a portrait photo. Ensure your entire face fits inside the frame and is well-lit. Supported formats: ',
     description: 'Instructions for portrait photo upload.',
   },
   'id.verification.camera.help.sight.question': {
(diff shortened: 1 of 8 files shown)


src/id-verification/IdVerification.messages.js
```diff
@@ -496,8 +496,13 @@ const messages = defineMessages({
     defaultMessage: 'Please upload an ID photo. Ensure the entire ID fits inside the frame and is well-lit. The file size must be under 10 MB. (Supported formats: .jpg, .jpeg, .png)',
     description: 'Instructions for ID photo upload.',
   },
-  'id.verification.id.photo.instructions.upload.error': {
-    id: 'id.verification.id.photo.instructions.upload.error',
+  'id.verification.id.photo.instructions.upload.error.invalidFileType': {
+    id: 'id.verification.id.photo.instructions.upload.error.invalidFileType',
+    defaultMessage: 'The file you have selected is not a supported image type. Please choose from the following formats: .jpg, .jpeg, .png',
```
schenedx, line 501: Let's also allow `bmp`, `gif`, `webp` and `tiff`

src/id-verification/ImageFileUpload.jsx
```diff
@@ -14,12 +18,15 @@ export default function ImageFileUpload({ onFileChange, intl }) {
     }
 
     const fileObject = e.target.files[0];
-    if (fileObject.size < maxFileSize) {
+    if (!['image/png', 'image/jpeg'].includes(fileObject.type)) {
```
schenedx, line 21: Please add more image types into this list. Also, I recommend you create a const variable to store this list in the function scope.
See the list at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

schenedx (changes requested):
Besides suggestion on expanding the accepted image types, the logic looks fine.

src/id-verification/ImageFileUpload.jsx
bseverino: I went ahead and pulled the supported image types out of the translation strings, so they'll be easier to update if needed (and they don't need to be translated anyway). I also changed the accept attribute back to a blanket `image/*` which I think should be fine, but I'll make sure the logging on the backend catches any errors related to this.

schenedx (approved):
👍

PR #425, opened 2021-04-05
feat: Allow users in experiment to continue with upload if no camera access

alangsto:
## [MST-716](https://openedx.atlassian.net/browse/MST-716)

If users are part of the A/B experiment for photo upload on the IDV flow, they should be allowed to continue with photo upload if they deny camera access or camera access is unsupported. If they have already denied camera access, they will not be allowed to switch between upload and photo capture mode throughout the flow. Instead the user will only be allowed to complete the flow by uploading photos from their device.

If a user is part of the experiment and has denied/unsupported camera access they will see this panel: 

Screen Shot 2021-04-05 at 1 38 19 PM

If they continue with upload, they will see this panel (that does not give them access to switch to photo capture mode):
Screen Shot 2021-04-05 at 1 38 29 PM

### DIFF
--- a/src/id-verification/IdVerification.messages.js
+++ b/src/id-verification/IdVerification.messages.js
@@ -11,6 +11,11 @@ const messages = defineMessages({
     defaultMessage: 'support',
     description: 'Website support.',
   },
+  'id.verification.continue.upload': {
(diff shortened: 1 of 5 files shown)


schenedx (approved):
👍

PR #426, opened 2021-04-06
[MST-717] Reroute user to summary panel in A/B test

bseverino:
[MST-717](https://openedx.atlassian.net/browse/MST-717)

The new reroutes introduced in the current A/B test override the summary panel reroute, resulting in incorrect behavior. This fix ensures that, if a user has reached the summary panel once, they are properly rerouted from the camera/upload panels after going back to retake or re-upload their photos.

### DIFF
--- a/src/id-verification/routing-utilities.js
+++ b/src/id-verification/routing-utilities.js
@@ -2,56 +2,79 @@ import { useContext } from 'react';
 import { useLocation } from 'react-router-dom';
 import IdVerificationContext, { MEDIA_ACCESS } from './IdVerificationContext';
 
+const SLUGS = {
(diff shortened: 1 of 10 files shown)


src/id-verification/routing-utilities.js
```diff
@@ -2,56 +2,79 @@ import { useContext } from 'react';
 import { useLocation } from 'react-router-dom';
 import IdVerificationContext, { MEDIA_ACCESS } from './IdVerificationContext';
 
+const SLUGS = {
```
Zacharis278, line 5: thanks for cleaning up these magic strings 💯

Zacharis278 (approved):

PR #430, opened 2021-04-15
Add additional post parameters to track photo modes

alangsto:
In addition to the other data being collected for the IDV experiment, we would also like to know exactly which mode a learner used to submit photos. This data is then used in a segment tracking event, as seen in this PR: https://github.com/edx/edx-platform/pull/27349


schenedx (approved):
👍

PR #431, opened 2021-05-03
fix: remove hard-coded edX from page title

davidjoy:
This uses variable substitution to insert SITE_NAME into index.html, rather than hard-coding “edX” into the file.  It also updates the .env files to use “Your Platform Name Here” as the default SITE_NAME.

If this looks correct, I can go make similar changes in other MFEs.

Jira: https://openedx.atlassian.net/browse/CENG-6
PR for Lilac: https://github.com/edx/frontend-app-account/pull/432


sarina (approved):
Do we have to set up configuration for this somewhere?

davidjoy:
I took a look in configuration, and the defaults there are "localhost" and an empty string, so I think we're good.  edx.org's configuration properly sets it to "edX" as well.

sarina:
LGTM. Be sure to make a PR into the lilac branch too

PR #433, opened 2021-05-06
fix: Resolving i18n violations, hard-coded edx.org-specific strings

davidjoy:
This is most easily reviewed commit-by-commit. 

You can read about it in the comments, but it includes a bit of a hack to show alternate messages if the SITE_NAME is set to 'edX'.  This is because we have no way of supplying site-specific strings today, but needed to make the defaults Open edX-friendly.

Later, if we can find a way to include site-specific stuff, we could undo this.  It's all annotated as TODOs.

Will cherry-pick to lilac.master once we get it in.

### DIFF
--- a/src/account-settings/delete-account/BeforeProceedingBanner.jsx
+++ b/src/account-settings/delete-account/BeforeProceedingBanner.jsx
@@ -6,6 +6,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Hyperlink } from '@edx/paragon';
 
 // Messages
+import { getConfig } from '@edx/frontend-platform';
 import messages from './messages';
 
 // Components
@@ -22,13 +23,14 @@ const BeforeProceedingBanner = (props) => {
       <FormattedMessage
         id="account.settings.delete.account.before.proceeding"
         defaultMessage="Before proceeding, please {actionLink}."
-        description="Error that appears if you are trying to delete your edX account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
+        description="Error that appears if you are trying to delete your {siteName} account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
         values={{
           actionLink: (
             <Hyperlink destination={supportArticleUrl}>
               {intl.formatMessage(messages[instructionMessageId])}
             </Hyperlink>
           ),
+          siteName: getConfig().SITE_NAME,
         }}
       />
     </Alert>
--- a/src/account-settings/delete-account/BeforeProceedingBanner.jsx
+++ b/src/account-settings/delete-account/BeforeProceedingBanner.jsx
@@ -23,7 +23,7 @@ const BeforeProceedingBanner = (props) => {
       <FormattedMessage
         id="account.settings.delete.account.before.proceeding"
         defaultMessage="Before proceeding, please {actionLink}."
-        description="Error that appears if you are trying to delete your {siteName} account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
+        description="Error that appears if you are trying to delete your account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
         values={{
           actionLink: (
             <Hyperlink destination={supportArticleUrl}>
--- a/src/account-settings/delete-account/PrintingInstructions.jsx
+++ b/src/account-settings/delete-account/PrintingInstructions.jsx
@@ -24,7 +24,7 @@ const PrintingInstructions = (props) => {
       <FormattedMessage
         id="account.settings.delete.account.text.3.edX"
         defaultMessage="You may also lose access to verified certificates and other program credentials like MicroMasters certificates. You can make a copy of these for your records before proceeding with deletion. {actionLink}."
-        description="A message in the user account deletion area warning users that deleting their account will prevent them from accessing their certificates."
+        description="A message in the user account deletion area warning users that deleting their account will prevent them from accessing their certificates. 'actionLink' is a HTML link with a full sentence that describes how to print a certificate."
         values={{ actionLink }}
       />
     );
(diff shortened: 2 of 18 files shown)


src/account-settings/delete-account/BeforeProceedingBanner.jsx
```diff
@@ -22,13 +23,14 @@ const BeforeProceedingBanner = (props) => {
       <FormattedMessage
         id="account.settings.delete.account.before.proceeding"
         defaultMessage="Before proceeding, please {actionLink}."
-        description="Error that appears if you are trying to delete your edX account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
+        description="Error that appears if you are trying to delete your {siteName} account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
```
sarina, line 26: do you need `{siteName}` in this description field? Elsewhere you use `platform`

src/account-settings/delete-account/PrintingInstructions.jsx
```diff
@@ -2,22 +2,39 @@ import React from 'react';
...
 import { Hyperlink } from '@edx/paragon';
 
+import { getConfig } from '@edx/frontend-platform';
 import messages from './messages';
 
 const PrintingInstructions = (props) => {
   const actionLink = (
     <Hyperlink
-      destination="https://edx.readthedocs.io/projects/edx-guide-for-students/en/latest/SFD_certificates.html#printing-a-certificate"
+      // TODO: What would a generic version of this link look like?  Should
```
sarina, line 11: Q: is there a RTD site for Open edX?

src/account-settings/delete-account/PrintingInstructions.jsx
```diff
@@ -2,22 +2,39 @@ import React from 'react';
...
   );
 
+  // TODO: We lack a good way of providing custom language for a particular site.  This is a hack
+  // to allow edx.org to mention MicroMasters certificates to fulfill its business requirements.
+  if (getConfig().SITE_NAME === 'edX') {
+    return (
+      <FormattedMessage
+        id="account.settings.delete.account.text.3.edX"
+        defaultMessage="You may also lose access to verified certificates and other program credentials like MicroMasters certificates. You can make a copy of these for your records before proceeding with deletion. {actionLink}."
+        description="A message in the user account deletion area warning users that deleting their account will prevent them from accessing their certificates."
```
sarina, line 27: Worth adding in some info like "actionLink is a full sentence that describes how to print a certificate"?

src/account-settings/delete-account/messages.js
```diff
@@ -13,22 +13,27 @@ const messages = defineMessages({
...
   },
   'account.settings.delete.account.text.2': {
     id: 'account.settings.delete.account.text.2',
-    defaultMessage: 'Once your account is deleted, you cannot use it to take courses on the edX app, edx.org, or any other site hosted by edX. This includes access to edx.org from your employer’s or university’s system and access to private sites offered by MIT Open Learning, Wharton Executive Education, and Harvard Medical School.',
+    defaultMessage: 'Once your account is deleted, you cannot use it to take courses on {siteName}.',
+    description: 'A message in the user account deletion area',
+  },
+  'account.settings.delete.account.text.2.edX': {
+    id: 'account.settings.delete.account.text.2.edX',
+    defaultMessage: 'Once your account is deleted, you cannot use it to take courses on the {siteName} app, edx.org, or any other site hosted by {siteName}. This includes access to edx.org from your employer’s or university’s system and access to private sites offered by MIT Open Learning, Wharton Executive Education, and Harvard Medical School.',
```
sarina, line 26: seems odd to abstract out edX in this sentence but 🤷‍♀️

src/account-settings/delete-account/messages.js
```diff
@@ -58,12 +63,17 @@ const messages = defineMessages({
...
   },
   'account.settings.delete.account.modal.text.2': {
     id: 'account.settings.delete.account.modal.text.2',
-    defaultMessage: 'If you proceed, you will be unable to use this account to take courses on the edX app, edx.org, or any other site hosted by edX. This includes access to edx.org from your employer\'s or university\'s system and access to private sites offered by MIT Open Learning, Wharton Executive Education, and Harvard Medical School.',
+    defaultMessage: 'If you proceed, you will be unable to use this account to take courses on {siteName}.',
+    description: 'Messaging in the dialog asking user to confirm that they want to delete their entire account',
+  },
+  'account.settings.delete.account.modal.text.2.edX': {
+    id: 'account.settings.delete.account.modal.text.2.edX',
+    defaultMessage: 'If you proceed, you will be unable to use this account to take courses on the {siteName} app, edx.org, or any other site hosted by {siteName}. This includes access to edx.org from your employer\'s or university\'s system and access to private sites offered by MIT Open Learning, Wharton Executive Education, and Harvard Medical School.',
```
sarina, line 76: also seems weird to abstract out edX. If you don't abstract it out, then we'll still have viable translations for this string, right?

src/account-settings/delete-account/BeforeProceedingBanner.jsx
davidjoy: Doh, that's an oversight. Oops.  Must have gotten overzealous with replacing strings in this file.

src/account-settings/delete-account/messages.js
davidjoy: I went back and forth... it seemed like if we had an enterprise site that used a different SITE_NAME it'd be more correct to use that, but yeah, agreed.  I figured it couldn't hurt, and reduced the number of "edX" strings in the repo. 😄

src/account-settings/delete-account/messages.js
davidjoy: Well, I figured I should make it right, and so these strings have new, edX-specific keys (lines 24/74).  If I recall how Transifex works, it may be that it'll automatically suggest existing translations if the string hasn't changed, even though the key has... if that's the case, I'll back out the `{siteName}` changes here.  Maybe I just will since they're not accomplishing much anyway.

src/account-settings/delete-account/PrintingInstructions.jsx
davidjoy: Not that I can find...

sarina (approved):

davidjoy:
This has been cherry-picked to lilac.master.

PR #435, opened 2021-05-17
chore: let renovate be more liberal about what it merges

davidjoy:
This repo was allowing both patch/minor updates of Paragon, but I've opened that up a bit with this PR.  This is based on similar config we've used in other repositories.

It loses the "stability days" that were here, but I don't think we need that for patch/minor updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,17 +1,15 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    ":automergeLinters",
+    ":automergeTesters",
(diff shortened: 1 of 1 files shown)


davidjoy:
@abutterworth I tagged you on this since it's changing the Paragon-mergin' stuff we were piloting in this repo.  Lemme know what you think or if you want to incorporate something from that back into this.

stvstnfrd (approved):

abutterworth (approved):

PR #440, opened 2021-06-02
build(ci): remove travis-ci in favor of github actions

stvstnfrd:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    - master
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+        - 12
+        npm-test:
```
stvstnfrd, line 15: Now with parallelization!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+        - lint
+        - test
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install -g npm@6
+    - run: make requirements
+    - run: make test NPM_TESTS=build
```
stvstnfrd, line 27: When running tests independently, always run `build` first.

Makefile
```diff
@@ -10,8 +10,19 @@ tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/$(transi
...
-requirements:
-	npm install
+NPM_TESTS=build i18n_extract lint test is-es5
+
+.PHONY: test
+test: $(addprefix test.npm.,$(NPM_TESTS))  ## validate ci suite
+
+.PHONY: test.npm.*
+test.npm.%: validate-no-uncommitted-package-lock-changes
+	test -d node_modules || $(MAKE) requirements
```
stvstnfrd, line 20: Note: I added this because it's helpful when running tests locally. In case a developer forgets to `make requirements`/`npm ci`/`npm install` for the first time, just do it automatically for them.

Makefile
```diff
@@ -10,8 +10,19 @@ tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/$(transi
 # This directory must match .babelrc .
 transifex_temp = ./temp/babel-plugin-react-intl
 
-requirements:
-	npm install
```
stvstnfrd, line 14: Any reason to _not_ always make this `npm ci` (for non-upgrade workflows)?

davidjoy: I think that makes sense to change.

Makefile
```diff
@@ -10,8 +10,19 @@ tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/$(transi
...
+NPM_TESTS=build i18n_extract lint test is-es5
+
+.PHONY: test
+test: $(addprefix test.npm.,$(NPM_TESTS))  ## validate ci suite
```
davidjoy, line 16: Not sure I understand what's happening here... is this just running `test.npm.*`?

davidjoy (approved):

Makefile
stvstnfrd: Yeah, it's taking a build list like ` NPM_TESTS=build i18n_extract lint test is-es5` and prefixing each item with `test.npm.` so that we don't have to tie the configuration to the "magic" named internal `make` target (`test.npm.X`).

PR #479, opened 2021-08-20
feat: Update IDV workflow to use the VerifiedName

schenedx:
If the Verified Name feature is turned on, and the user do have a VerifiedName, use that name on the Account Name Check page of the IDV flow instead of account profile name

### DIFF
--- a/src/account-settings/data/service.js
+++ b/src/account-settings/data/service.js
@@ -176,12 +176,12 @@ export async function shouldDisplayDemographicsQuestions() {
   return false;
 }
 
-export async function getVerifiedName(username) {
+export async function getVerifiedName() {
   let data;
   const client = getAuthenticatedHttpClient();
   try {
-    ({ data } = await client
-      .get(`${getConfig().LMS_BASE_URL}/api/edx_name_affirmation/v1/verified_name?username=${username}`));
+    const requestUrl = `${getConfig().LMS_BASE_URL}/api/edx_name_affirmation/v1/verified_name`;
+    ({ data } = await client.get(requestUrl));
   } catch (error) {
     return {};
   }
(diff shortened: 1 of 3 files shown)


src/account-settings/data/service.js
```diff
@@ -215,7 +218,7 @@ export async function getSettings(username, userRoles, userId) {
     getConfig().ENABLE_DEMOGRAPHICS_COLLECTION && shouldDisplayDemographicsQuestions(),
     getConfig().ENABLE_DEMOGRAPHICS_COLLECTION && getDemographics(userId),
     getConfig().ENABLE_DEMOGRAPHICS_COLLECTION && getDemographicsOptions(),
-    getVerifiedName(username),
+    getVerifiedName(),
```
alangsto, line 221: Is `getVerifiedName` ever supplied a username? The function is expecting that arg, but it's never being supplied here. Can we just remove that parameter from the function?

Zacharis278:
this is somewhat high-level but this PR gets me thinking. What happens if the platform does not have the name-affirmation plugin installed?

src/account-settings/data/service.js
schenedx: Yeah, removing the parameter make sense in this PR. I'll make that change.

schenedx:
@Zacharis278 I'd hope that the `||` operator would just serve the purpose of being flexible. What do you have in mind to be explicit? Waffle? frontend config?

src/account-settings/data/service.js
schenedx: @alangsto Took in your feedback. Please review again.

alangsto (approved):

PR #482, opened 2021-08-23
fix: Should not display error message when the account name matches the ID name

schenedx:
### Description
MST-990, we should fix the problem where the ID name error message still show up with Account name filled in or Verified name filled in.

### Before the change:
beforeIdNameErrorFix

### After the change:
afterIDNameErrorFix


alangsto (approved):
Pending lint fixes this looks good to me!

PR #487, opened 2021-08-28
feat: Make VerifiedName not editable and show proper icon when the Name is in pending status

schenedx:
### Description

[MST-957](https://openedx.atlassian.net/browse/MST-957)

When the VerifiedName is in pending status, 
update the editable field for VerifiedName to be not editable,
and show yellow warning icon. See screenshot below

pendingVerification

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -14,7 +14,7 @@ import {
   getLanguageList,
 } from '@edx/frontend-platform/i18n';
 import { Hyperlink, Icon } from '@edx/paragon';
-import { CheckCircle } from '@edx/paragon/icons';
+import { CheckCircle, WarningFilled } from '@edx/paragon/icons';
 
 import messages from './AccountSettingsPage.messages';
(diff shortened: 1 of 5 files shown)


bseverino (approved):

MichaelRoytman:
It looks good to me. We did want the pending names to be grayed out in this state. Is that easy for you to add?

schenedx:
@MichaelRoytman I need to dive into that a bit. Good call out tho.

schenedx:
@MichaelRoytman Added grayout toggle to the EditableField component. See screenshot:

grayedoutverifiedname

MichaelRoytman (approved):
The change to the grayed text for verified name looks good to me.

One thing I notice is that the text of the profile name should also be grayed out and should also be missing its edit button. It should also have the text `Your public facing name that appears on your certificates. Full Name cannot be changed while identity verification is pending.`.

If you want to create another ticket for that work or follow up in a new PR, that's fine with me.

schenedx:
@MichaelRoytman Hmm.. I am not fully clear on why with a pending verifiedName, we need to also disable the update of profile_name. 
Can we please talk about it later today?

PR #490, opened 2021-08-30
feat: Add Dismissible Alert Informing Learner of Failed Name Verification

MichaelRoytman:
If a learner requests a profile name change or verified name change that requires ID verification, their ID verification submission may be rejected. This code change adds a dismissible error alert to the Account Settings page informing the learner of the failure and redirecting them to a help article about ID verification failures. The local storage key used to determine whether to show the alert or not uses the created time of the name verification object to ensure that subsequent ID verification failures still trigger this alert, even for the same name.

MST-954 (https://openedx.atlassian.net/browse/MST-954)

image

I'll be working on a code change to revert to restoring the previous name affirmation state.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -13,8 +13,8 @@ import {
   getCountryList,
   getLanguageList,
 } from '@edx/frontend-platform/i18n';
-import { Hyperlink, Icon } from '@edx/paragon';
-import { CheckCircle, WarningFilled } from '@edx/paragon/icons';
+import { Button, Hyperlink, Icon } from '@edx/paragon';
+import { CheckCircle, Error, WarningFilled } from '@edx/paragon/icons';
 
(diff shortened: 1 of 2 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -228,8 +228,43 @@ class AccountSettingsPage extends React.Component {
         id="dismissedVerifiedNameSuccessMessage"
         variant="success"
         icon={CheckCircle}
-        header={this.props.intl.formatMessage(messages['account.settings.field.name.verified.sucess.message.header'])}
-        body={this.props.intl.formatMessage(messages['account.settings.field.name.verified.sucess.message'])}
+        header={this.props.intl.formatMessage(messages['account.settings.field.name.verified.success.message.header'])}
```
MichaelRoytman, line 231: The change here is `sucess` to `success`.

schenedx (approved):
👍

PR #491, opened 2021-08-30
feat: Change Which VerifiedName Record is Displayed When Most Recent VerifiedName is Denied

MichaelRoytman:
If a learner requests a profile name change or verified name change that requires ID verification, their ID verification submission may be rejected. This code change changes the verified name that is displayed in this case. If a learner's most recent verified name record has been denied, and the learner has a previously approved verified name, display that. If If a learner's most recent verified name record has been denied, and the learner either has no previously verified name records or no previous verified name records at all, do not show a verified name field.

It also refactors the code to better support UI treatments for other verified name statuses, like pending and submitted.
MST-954 (https://openedx.atlassian.net/browse/MST-954)

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -138,18 +138,19 @@ class AccountSettingsPage extends React.Component {
     })),
   }));
 
-  getMostRecentVerifiedNameRecord = verifiedNameHistory => {
-    if (verifiedNameHistory.length > 0) {
-      return verifiedNameHistory.reduce(
-        (prev, curr) => {
-          const prevDate = new Date(prev.created);
-          const currDate = new Date(curr.created);
-          return currDate > prevDate ? curr : prev;
-        },
-      );
+  sortDates = (a, b) => {
+    const aTimeSinceEpoch = new Date(a).getTime();
+    const bTimeSinceEpoch = new Date(b).getTime();
+
+    return bTimeSinceEpoch - aTimeSinceEpoch;
+  }
+
+  sortVerifiedNameRecords = verifiedNameHistory => {
+    if (Array.isArray(verifiedNameHistory)) {
+      return [...verifiedNameHistory].sort(this.sortDates);
     }
 
-    return null;
+    return [];
   }
 
   handleEditableFieldChange = (name, value) => {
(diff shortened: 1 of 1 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -138,18 +138,19 @@ class AccountSettingsPage extends React.Component {
...
+  sortDates = (a, b) => {
+    const aTimeSinceEpoch = new Date(a).getTime();
+    const bTimeSinceEpoch = new Date(b).getTime();
+
+    return bTimeSinceEpoch - aTimeSinceEpoch;
+  }
+
+  sortVerifiedNameRecords = verifiedNameHistory => {
+    if (Array.isArray(verifiedNameHistory)) {
+      return [...verifiedNameHistory].sort(this.sortDates);
```
MichaelRoytman, line 150: `Array.prototype.sort` is one of the few Javascript methods that modifies arrays in place, so I have spread the array here to avoid modifying the API response data.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -354,7 +389,7 @@ class AccountSettingsPage extends React.Component {
     return (
       <>
         <div className="account-section" id="basic-information" ref={this.navLinkRefs['#basic-information']}>
-          {showVerifiedName && verifiedNameMessage}
+          {verifiedNameEnabled && this.renderVerifiedNameMessage(mostRecentVerifiedName)}
```
MichaelRoytman, line 392: I had to change this because the conditions for rendering the alert and for rendering the field have diverged.

alert: if verifiedNameEnabled and mostRecentVerifiedName has an alert treatment
field: if verifiedNameEnabled and there exists an approvedVerifiedName

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -271,6 +270,41 @@ class AccountSettingsPage extends React.Component {
...
+      case 'approved':
+        return this.renderVerifiedNameSuccessMessage();
+      case 'denied':
+        return this.renderVerifiedNameFailureMessage(verifiedName, created);
+      default:
+        return null;
+    }
+  }
+
+  renderVerifiedNameIcon = (status) => {
```
alangsto, line 286: Thank you for fixing this!

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -326,23 +360,24 @@ class AccountSettingsPage extends React.Component {
...
 
-    let verifiedNameMessage;
-    let showVerifiedApproved = false;
-    switch (verifiedName.status) {
+    let verifiedNameToDisplay = null;
+    switch (mostRecentVerifiedName && mostRecentVerifiedName.status) {
       case 'approved':
-        showVerifiedApproved = true;
-        verifiedNameMessage = this.renderVerifiedNameSuccessMessage();
-        break;
```
alangsto, line 339: Don't we still want to define a `verifiedNameToDisplay` if the status is approved?

alangsto: Ahh got it, I didn't realize that. Thanks for explaining!

alangsto (approved):
LGTM

PR #492, opened 2021-08-31
feat: add alert and correct messaging for submitted verified name

alangsto:
## [MST-956](https://openedx.atlassian.net/browse/MST-956)

If a learner has a submitted verified name change, we should alert them of that submitted name change. In addition to displaying an alert, we also want to make both the verified name and full name fields grayed-out and uneditable. The help text for each of those fields should also be updated.

Screen Shot 2021-08-31 at 3 35 11 PM

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -13,14 +13,15 @@ import {
   getCountryList,
   getLanguageList,
 } from '@edx/frontend-platform/i18n';
-import { Button, Hyperlink, Icon } from '@edx/paragon';
+import {
+  Button, Hyperlink, Icon, Alert,
+} from '@edx/paragon';
 import { CheckCircle, Error, WarningFilled } from '@edx/paragon/icons';
 
 import messages from './AccountSettingsPage.messages';
 import { fetchSettings, saveSettings, updateDraft } from './data/actions';
 import { accountSettingsPageSelector } from './data/selectors';
 import PageLoading from './PageLoading';
-import Alert from './Alert';
 import JumpNav from './JumpNav';
 import DeleteAccount from './delete-account';
 import EditableField from './EditableField';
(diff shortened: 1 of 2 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -180,7 +181,7 @@ class AccountSettingsPage extends React.Component {
 
     return (
       <div>
-        <Alert className="alert alert-danger" role="alert">
+        <Alert variant="danger">
```
alangsto, line 184: Updated to use the paragon Alert component, verified the components looks the same

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -305,6 +322,15 @@ class AccountSettingsPage extends React.Component {
     }
   }
 
+  renderFullNameHelpText = (status) => {
+    switch (status) {
+      case 'pending':
+        return (this.props.intl.formatMessage(messages['account.settings.field.name.full.help.text.pending']));
```
schenedx, line 328: Why is this message location reversed with the message location of line 330? This line has `field.name.full`. The line 330 one has `field.full.name`.

schenedx (approved):
Just a minor comment. The change LGTM.

src/account-settings/AccountSettingsPage.messages.jsx
```diff
@@ -106,6 +106,11 @@ const messages = defineMessages({
     defaultMessage: 'Verification Pending. This usually takes 48 hours or less. Verified Name cannot be changed while identity verification is pending',
     description: 'Help text for the account settings verified name field when the name is pending verification.',
   },
+  'account.settings.field.full.name.help.text.pending': {
+    id: 'account.settings.field.full.name.help.text.pending',
+    defaultMessage: 'Your public facing name that appears on your certificates. Full Name cannot be changed while identity verification is pending.',
```
MichaelRoytman, line 111: Nit: Since the label for full name is "Full name", I think we should capitalize it the same way in the help text.

PR #493, opened 2021-09-03
fix: Fix Broken Editing of Verified Name Field

MichaelRoytman:
This code change fixes a bug where the verified name field was not editable. This was due to the fact that the verified name was not wired up to the Redux store, so draft verified names were not stored in the Redux store.

### DIFF
--- a/src/account-settings/data/selectors.js
+++ b/src/account-settings/data/selectors.js
@@ -7,9 +7,63 @@ export const accountSettingsSelector = state => ({ ...state[storeName] });
 
 const editableFieldNameSelector = (state, props) => props.name;
 
+const sortedVerifiedNameHistorySelector = createSelector(
+  accountSettingsSelector,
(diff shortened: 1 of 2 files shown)


src/account-settings/data/selectors.js
```diff
@@ -69,7 +123,18 @@ export const profileDataManagerSelector = createSelector(
 
 export const staticFieldsSelector = createSelector(
   accountSettingsSelector,
-  accountSettings => (accountSettings.profileDataManager ? ['name', 'email', 'country'] : []),
+  mostRecentVerifiedNameSelector,
```
MichaelRoytman, line 126: I was on the fence about this. I imagine staticFields itself is meant to be static and not dynamic, but I kept it this way because `isEditable` leverages this state.

bseverino (approved):
LGTM 👍

PR #500, opened 2021-09-13
feat: if verified name is on, we can redo ID verification

ashultz0:
The verified_name endpoint provides is_enabled only if a name exists and otherwise 404s, so we have to request the enabled flag separately through the new endpoint to make sure we get it.

Validated on local build with an approved IDV and name affirmation on and off.

MST-1026

### DIFF
--- a/src/id-verification/IdVerificationContextProvider.jsx
+++ b/src/id-verification/IdVerificationContextProvider.jsx
@@ -2,7 +2,7 @@ import React, { useState, useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { AppContext } from '@edx/frontend-platform/react';
 
-import { getProfileDataManager, getVerifiedName } from '../account-settings/data/service';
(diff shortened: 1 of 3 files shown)


src/id-verification/IdVerificationContextProvider.jsx
```diff
@@ -30,11 +30,25 @@ export default function IdVerificationContextProvider({ children }) {
     hasGetUserMediaSupport ? MEDIA_ACCESS.PENDING : MEDIA_ACCESS.UNSUPPORTED,
   );
 
+  const [verifiedNameEnabled, setVerifiedNameEnabled] = useState('');
```
bseverino, line 33: NIT: I would go with a default of false here instead of an empty string, to keep the type consistent - everything else LGTM 👍

bseverino (approved):

src/id-verification/IdVerificationContextProvider.jsx
ashultz0: 5640fb9

PR #505, opened 2021-09-21
feat: include pending or approved verified names in the name field if they exist

MichaelRoytman:
[MST-1016](https://openedx.atlassian.net/browse/MST-1016)

If a learner has a pending or approved verified name, the most recent should take precedence over the profile name on the GetNameIdPanel during the "Account Name Check". If the learner has no pending or approved verified names, then the learner's profile name should be used instead.

### DIFF
--- a/src/id-verification/VerifiedNameContext.jsx
+++ b/src/id-verification/VerifiedNameContext.jsx
@@ -1,0 +1,40 @@
+import React, { createContext, useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+
+import { getVerifiedNameHistory } from '../account-settings/data/service';
+import { getMostRecentApprovedOrPendingVerifiedName } from '../utils';
+
+export const VerifiedNameContext = createContext();
+
+export function VerifiedNameContextProvider({ children }) {
+  const [verifiedNameEnabled, setVerifiedNameEnabled] = useState(false);
+  const [verifiedName, setVerifiedName] = useState('');
+  useEffect(() => {
+    // Make API call to retrieve VerifiedName history for the learner.
+    // From this information, derive whether the verified name feature is enabled
+    // and the learner's verified name as it should be displayed during the IDV process.
+    (async () => {
+      const response = await getVerifiedNameHistory();
+      if (response) {
+        const { verified_name_enabled: verifiedNameFeatureEnabled, results } = response;
+        setVerifiedNameEnabled(verifiedNameFeatureEnabled);
+
+        if (verifiedNameFeatureEnabled) {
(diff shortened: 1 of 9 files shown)


src/utils.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Compare two dates.
+ * @param {*} a the first date
+ * @param {*} b the second date
+ * @returns a negative integer if a > b, a positive integer if a < b, or 0 if a = b
+ */
+export function compareVerifiedNamesByCreatedDate(a, b) {
```
MichaelRoytman, line 7: Please also be aware that I fixed a bug where the dates were not being instantiated using the `created` field. This wasn't just a simple move of the function.

src/id-verification/VerifiedNameContext.jsx
```diff
@@ -0,0 +1,40 @@
...
+        const { verified_name_enabled: verifiedNameFeatureEnabled, results } = response;
+        setVerifiedNameEnabled(verifiedNameFeatureEnabled);
+
+        if (verifiedNameFeatureEnabled) {
+          const applicableVerifiedName = getMostRecentApprovedOrPendingVerifiedName(results);
+          setVerifiedName(applicableVerifiedName);
+        }
+      }
+    })();
+  }, []);
```
schenedx, line 28: This means the useEffect will be called only once during the page load. Is that intended?

schenedx (approved):
👍

bseverino (approved):

src/id-verification/VerifiedNameContext.jsx
MichaelRoytman: I think so. The originals calls to `getVerifiedName` and `getVerifiedNameEnabled` that the new call to `getVerifiedNameHistory` replaces behaved the same way.

I cannot think of a scenario during the IDV process where we'd want to pull in new data. The data returned by the API shouldn't change based on any actions the learner takes during the IDV process (before submission).

Do you have concerns about the state of this data changing on the server while the learner is in the middle of IDV? I think handling those cases would be pretty tricky. For example, if the flag is turned on or off or a VerifiedName moves from the submitted to the approved status while the learner is on the ready to submit panel, I'm not sure what the best course of action would be. That would be better to explore in a follow on ticket if we want to do that, but I fear it may be hairy.

schenedx: @MichaelRoytman Yeah, agree the scenario you described is too rare to worry right now.
I just want to make you aware a recent learning I had. I think this is good. And we shall find out what we cannot anticipate.

PR #506, opened 2021-09-23
Fix i18n issues

sarina:
(1) Indicate to translators when strings that reference `edX` are only being used on edx.org. Also add notes about what _shouldn't_ be translated.

(2) Fix an issue where strings were split into two messages. This causes horrible issues with RTL languages, and looks pretty bad in any language that doesn't exactly follow English semantics (ie, most)

### DIFF
--- a/src/account-settings/delete-account/BeforeProceedingBanner.jsx
+++ b/src/account-settings/delete-account/BeforeProceedingBanner.jsx
@@ -21,15 +21,10 @@ const BeforeProceedingBanner = (props) => {
       icon={<FontAwesomeIcon className="mr-2" icon={faExclamationTriangle} />}
     >
       <FormattedMessage
-        id="account.settings.delete.account.before.proceeding"
-        defaultMessage="Before proceeding, please {actionLink}."
-        description="Error that appears if you are trying to delete your account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
+        defaultMessage=intl.formatMessage(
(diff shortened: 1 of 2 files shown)


src/account-settings/delete-account/BeforeProceedingBanner.jsx
```diff
@@ -21,15 +21,10 @@ const BeforeProceedingBanner = (props) => {
       icon={<FontAwesomeIcon className="mr-2" icon={faExclamationTriangle} />}
     >
       <FormattedMessage
-        id="account.settings.delete.account.before.proceeding"
-        defaultMessage="Before proceeding, please {actionLink}."
-        description="Error that appears if you are trying to delete your account, but something about your account needs attention first.  The actionLink will be instructions, such as 'unlink your Facebook account'."
+        defaultMessage=intl.formatMessage(
+          messages[instructionMessageId], {linkStart: <Hyperlink destination={supportArticleURL}>, linkEnd=</Hyperlink>}
```
davidjoy, line 25: Yeah, I _think_ this is unfortunately problematic.  I'm fairly certain that `linkStart` in this case will get cast to a string and become `[Object object]` instead of the actual HTML content.  That's the difference between `formatMessage` and `FormattedMessage`. 

The other problem is that a closing tag on its own is a syntax error for the JSX parser; they must be properly paired for JSX parsing to work. 😬 

Ugh, trying to figure out how to do this right... I believe this is why it ended up this way. 🤔

sarina:
I'm going to close this for now and await the ability to update react-intl

PR #507, opened 2021-09-24
submit verified name change

Zacharis278:
[MST-1029](https://openedx.atlassian.net/browse/MST-1029)

1. Update the NameChange modal to function with both the profile and verified name fields. The UX is effectively the same for updating either name, if you enter verified name into the form it will just prefill the input on the last panel.

2. Was running into some issues were verifiedName in the form was bound to the verifiedName model object. This was causing updating the form to override the object with a string. I did some refactoring to move state values that aren't reflected in the form itself out of the formData object.

There's one functional change as a result of this that could be up for discussion. If the user has a previously approved verified name that name gets pre-populated into the form when changing your profile name. It certainly doesn't have to happen that way, but left it for now.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -476,7 +476,7 @@ class AccountSettingsPage extends React.Component {
             type="text"
             value={
               verifiedNameEnabled
-              && verifiedName.status === 'submitted'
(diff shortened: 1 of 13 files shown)


bseverino (approved):
LGTM. I think it makes sense to pre-populate the verified name field if they already have one 👍

PR #514, opened 2021-09-28
fix: Display Last Approved Verified Name When Most Recent Verified Name is Pending

MichaelRoytman:
[MST-1073](https://openedx.atlassian.net/browse/MST-1073)

When a learner's most recent Verified Name is in the pending state, we should display their most recent approved Verified Name. If such a Verified Name does not exist, do not display any Verified Name. In either case, do not disable or gray out either the Full name or Verified name fields.

This change is being made because we do not want to prevent a learner from editing the Verified name field if their most recent entry is pending. An entry remains in the pending state when a learner has created a Verified name by changing their name on the Account Settings page but has not followed through with submitting their IDV.


Zacharis278 (approved):
LGTM

schenedx (approved):
👍

PR #515, opened 2021-09-29
fix: Checkbox incorrectly tied to profile name

alangsto:
Fixed a bug that incorrectly rendered the certificate preference checkbox upon each page render. The bug caused 
* the checkboxes to both be selected at the same time
* the checkbox to always be selected for the profile name
* wouldn't allow the proper UI changes upon saving a verified name (i.e. save button wouldn't have a spinner, form wouldn't close)

The checkbox selection works correctly now, and the save button for the verified name field also works as expected.

### DIFF
--- a/src/account-settings/certificate-preference/CertificatePreference.jsx
+++ b/src/account-settings/certificate-preference/CertificatePreference.jsx
@@ -76,7 +76,7 @@ function CertificatePreference({
   }
 
   useEffect(() => {
-    if (fieldName === 'verifiedName') {
+    if (fieldName === 'verified_name') {
       setChecked(useVerifiedNameForCerts);
     } else {
       setChecked(!useVerifiedNameForCerts);
--- a/src/account-settings/data/selectors.js
+++ b/src/account-settings/data/selectors.js
@@ -63,7 +63,7 @@ const valuesSelector = createSelector(
   mostRecentApprovedVerifiedNameValueSelector,
   (accountSettings, mostRecentApprovedVerifiedNameValue) => {
     let useVerifiedNameForCerts = (
-      accountSettings.values.verifiedNameHistory?.use_verified_name_for_certs || false
+      accountSettings.verifiedNameHistory?.use_verified_name_for_certs || false
     );
 
     if (Object.keys(accountSettings.confirmationValues).includes('useVerifiedNameForCerts')) {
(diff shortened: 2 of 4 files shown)


src/account-settings/data/selectors.js
```diff
@@ -63,7 +63,7 @@ const valuesSelector = createSelector(
   mostRecentApprovedVerifiedNameValueSelector,
   (accountSettings, mostRecentApprovedVerifiedNameValue) => {
     let useVerifiedNameForCerts = (
-      accountSettings.values.verifiedNameHistory?.use_verified_name_for_certs || false
+      accountSettings?.verifiedNameHistory.use_verified_name_for_certs || false
```
schenedx, line 66: What does the `?` right before the `.` mean?
I never seen this in javascript. Is this typescript syntax?

src/account-settings/certificate-preference/CertificatePreference.jsx
```diff
@@ -76,7 +76,7 @@ function CertificatePreference({
   }
 
   useEffect(() => {
-    if (fieldName === 'verifiedName') {
+    if (fieldName === 'verified_name') {
```
schenedx, line 79: Is there an easy way to write a unit test covering this bug here?

src/account-settings/data/selectors.js
schenedx: Also, should we have a unit test to cover this change?

schenedx (commented):
Just call for unit tests.

src/account-settings/data/selectors.js
bseverino: Optional chaining: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

However in this case I think we should keep the operator on `verifiedNameHistory`, in case it returns null

alangsto: Noted this in slack, but will also copy here: 

Zach and I chatted about adding testing for the selectors, and determined that it wouldn't be worth it to add it for this ticket. This is because:
* If we were to add end to end testing (which is what is typically done to test selectors/redux), we could update the tests for the CertificatePreference by mocking out the store, not the selector. If we do that, any time someone updates the selector code, or adds a new form field, the CertificatePreference code will break. This seems brittle.
* If we want to test the redux aspect, testing for this is not typically done on individual pieces (i.e. a test suite only for selectors), and instead is done on the saga/reducer/selector combination. Doing this would require creating a test structure from scratch, which seems out of scope for this ticket.

schenedx (approved):
👍

PR #516, opened 2021-09-29
fix: approved alert should appear for subsequent approved attempts

alangsto:
## [MST-1072](https://openedx.atlassian.net/browse/MST-1072)

Current behavior:
- Learner changes their name, is prompted to complete IDV, and it is approved
- Learner dismisses the "approved" alert on their account settings page
- Learner submits another name change, and it is approved
- The "approved" alert will not appear again on the account settings page, because the local storage value determining whether the alert has been dismissed still persists in the browser

Behavior with fix:
- Learner changes their name, is prompted to complete IDV, and it is approved
- Learner dismisses the "approved" alert on their account settings page
- Learner submits another name change, and it is approved
- New "approved" alert is visible on their account settings page

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -278,15 +278,20 @@ class AccountSettingsPage extends React.Component {
     }
   }
 
-  renderVerifiedNameSuccessMessage = () => (
-    <OneTimeDismissibleAlert
-      id="dismissedVerifiedNameSuccessMessage"
-      variant="success"
-      icon={CheckCircle}
-      header={this.props.intl.formatMessage(messages['account.settings.field.name.verified.success.message.header'])}
-      body={this.props.intl.formatMessage(messages['account.settings.field.name.verified.success.message'])}
-    />
-  )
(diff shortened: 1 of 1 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -278,15 +278,20 @@ class AccountSettingsPage extends React.Component {
...
-  renderVerifiedNameSuccessMessage = () => (
-    <OneTimeDismissibleAlert
-      id="dismissedVerifiedNameSuccessMessage"
-      variant="success"
-      icon={CheckCircle}
-      header={this.props.intl.formatMessage(messages['account.settings.field.name.verified.success.message.header'])}
-      body={this.props.intl.formatMessage(messages['account.settings.field.name.verified.success.message'])}
-    />
-  )
+  renderVerifiedNameSuccessMessage = (verifiedName, created) => {
```
alangsto, line 281: I could simplify code between `renderVerifiedNameSuccessMessage` and `renderVerifiedNameFailureMessage`, as some of the code is duplicated, but wasn't sure if I should follow the pattern of each status having its own message function.

bseverino (approved):

PR #517, opened 2021-09-29
feat: add warning helptext to name fields after verification failure

Zacharis278:
Adds warnings to the profile and verified name fields if that last verification failed. These warnings are tied to dismissing the failure notification since otherwise they would persist until the user is verified.

Some of the messaging on this may change since the UI doesn't know what triggered the verification that failed. It may not be a name change is is implied. Meeting w/ Gabe tomorrow to discuss.

@edx/masters-devs-cosmonauts

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -153,6 +153,12 @@ class AccountSettingsPage extends React.Component {
     return bTimeSinceEpoch - aTimeSinceEpoch;
   }
 
+  verificationFailureAcked = verifiedNameObj => (
+    localStorage.getItem(
+      `dismissedVerifiedNameFailureMessage-${verifiedNameObj.verified_name}-${new Date(verifiedNameObj.created).valueOf()}`,
+    ) === 'true'
+  )
+
   sortVerifiedNameRecords = verifiedNameHistory => {
     if (Array.isArray(verifiedNameHistory)) {
       return [...verifiedNameHistory].sort(this.sortDates);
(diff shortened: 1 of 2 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -270,6 +277,18 @@ class AccountSettingsPage extends React.Component {
     switch (status) {
       case 'submitted':
         return this.props.intl.formatMessage(messages['account.settings.field.full.name.help.text.submitted']);
+      case 'denied':
+        if (this.props.committedValues.name !== profileName && !this.verificationFailureAcked(verifiedNameObj)) {
```
Zacharis278, line 281: `this.props.committedValues.name !== profileName` is a bit of a jank to loosely determine if the verification was the result of a profile name change but it's a bit brittle.

schenedx: Could you add some comments in the code to clarify the reason for the if check?

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -153,6 +153,12 @@ class AccountSettingsPage extends React.Component {
     return bTimeSinceEpoch - aTimeSinceEpoch;
   }
 
+  verificationFailureAcked = verifiedNameObj => (
+    localStorage.getItem(
+      `dismissedVerifiedNameFailureMessage-${verifiedNameObj.verified_name}-${new Date(verifiedNameObj.created).valueOf()}`,
```
schenedx, line 158: Could you look into how this name is constructed and saved into local storage? I am wondering if we can refactor that into an util function so the logic is consistent on `get` and `set`

schenedx (approved):
Minor suggestions inline. LGTM

src/account-settings/AccountSettingsPage.jsx
Zacharis278: good idea

PR #520, opened 2021-10-01
MST-1065: Add Copy and UI Treatment for Name Affirmation When Enabled 

MichaelRoytman:
[MST-1065](https://openedx.atlassian.net/browse/MST-1065)

When the Name Affirmation feature is enabled, a few panels in the IDV workflow should have different copy to reflect the changes associated with Name Affirmation. This is because entering a new during the IDV flow (GetNameIdPanel) will no longer modify the learner's "Account Name". Furthermore, "Account Name" is an ambiguous term; it's not clear that it refers to the learner's full name on their profile. These copy changes make the panel more generic and remove details about name changes. The SummaryPanel also has a minor copy change to change "Account Name" to "Name".

These code changes also include some UI changes.

**Before:**
image
image

**After:**

image
image

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 const { createConfig } = require('@edx/frontend-build');
 
 module.exports = createConfig('jest', {
-  setupFiles: [
+  setupFilesAfterEnv: [
     '<rootDir>/src/setupTest.js',
   ],
 });
--- a/src/id-verification/_id-verification.scss
+++ b/src/id-verification/_id-verification.scss
@@ -14,14 +14,6 @@
         max-height: 10rem;
       }
     }
-    .form-check {
-      padding: 0.5rem 0.5rem 1rem;
-      .form-check-label {
-        margin-left: 0.5rem;
-        padding-top: 0.2rem;
-        font-weight: normal;
-      }
-    }
   }
   .action-row {
     display: flex;
--- a/src/id-verification/tests/AccessBlocked.test.jsx
+++ b/src/id-verification/tests/AccessBlocked.test.jsx
@@ -4,7 +4,6 @@ import { createMemoryHistory } from 'history';
 import {
   render, cleanup, act, screen,
 } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
 import '@edx/frontend-platform/analytics';
 import { injectIntl, IntlProvider } from '@edx/frontend-platform/i18n';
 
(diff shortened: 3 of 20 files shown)


davidjoy:
(remember to use conventional commits when squashing/merging this!)

jest.config.js
```diff
@@ -1,7 +1,7 @@
 const { createConfig } = require('@edx/frontend-build');
 
 module.exports = createConfig('jest', {
-  setupFiles: [
+  setupFilesAfterEnv: [
```
MichaelRoytman, line 4: This is the recommended setup from jest-dom's [docs](https://github.com/testing-library/jest-dom/tree/v5.14.0#usage).

`expect` isn't defined if this isn't changed to `setupFilesAfterEnv` from `setupFiles`.

Here is an explanation of the difference ([source](https://jestjs.io/docs/configuration#setupfilesafterenv-array)).

> Since setupFiles executes before the test framework is installed in the environment, this script file presents you the opportunity of running some code immediately after the test framework has been installed in the environment.

src/id-verification/_id-verification.scss
```diff
@@ -14,14 +14,6 @@
         max-height: 10rem;
       }
     }
-    .form-check {
```
MichaelRoytman, line 17: I removed this SCSS; it was messing up the default styles of `Form.Check` in the `GetNameIdPanel` when the `Form.Check` isn't stacked.

This SCSS was used in two other places. I've detailed the visual changes below. Let me know if you feel they're significant.

Camera Before:

image

Camera After:

image


Chose Mode Before:

image


Choose Mode After:

image

src/id-verification/tests/panels/GetNameIdPanel.test.jsx
```diff
@@ -56,25 +63,26 @@ describe('GetNameIdPanel', () => {
     const nextButton = await screen.findByTestId('next-button');
     const errorMessageQuery = await screen.queryByTestId('id-name-feedback-message');
 
-    expect(input).toHaveProperty('readOnly');
```
MichaelRoytman, line 59: My understanding is that `toHaveProperty` is meant to check properties on an object, not attributes on HTML elements.

src/setupTest.js
```diff
@@ -1,5 +1,6 @@
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
+import '@testing-library/jest-dom';
```
MichaelRoytman, line 3: This is the recommended setup from jest-dom's [docs](https://github.com/testing-library/jest-dom/tree/v5.14.0#usage).

This means we can remove the `import '@testing-library/jest-dom/extend-expect';` lines from test files.

src/id-verification/tests/AccessBlocked.test.jsx
```diff
@@ -4,7 +4,6 @@ import { createMemoryHistory } from 'history';
 import {
   render, cleanup, act, screen,
 } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
```
MichaelRoytman, line 7: See my comment on `src/setupTest.js`.

schenedx (approved):
👍

PR #524, opened 2021-10-04
[MST-1099] Fix Flickering in IDV Panel Loading and Inconsistent Panel Loading

MichaelRoytman:
[MST-1099](https://openedx.atlassian.net/browse/MST-1099)

Due to a bug in the way the canVerify React state is being set, when the verified name feature is enabled, the code inconsistently loads either the AccessBlocked panel or the ReviewRequirementsPanel when the learner should be allowed to verify. If the learner is ineligible to complete IDV due to their IDV history (e.g. they already have an approved IDV), and the verified name feature is enabled, the learner should be able to verify and should not see the AccessBlocked panel. This code change fixes this bug.

This code change also fixes flickering that occurs when IDV is loading. This code change also includes some copy fixes.

### DIFF
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -1,0 +1,23 @@
+import { useEffect, useState } from 'react';
+
+// eslint-disable-next-line import/prefer-default-export
+export function useAsyncCall(asyncFunc) {
+  const [isLoading, setIsLoading] = useState();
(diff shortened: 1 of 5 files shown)


alangsto (approved):
LGTM!

PR #527, opened 2021-10-08
[MST-1104] Redirect user to original location after IDV

bseverino:
[MST-1104](https://openedx.atlassian.net/browse/MST-1104)

Allow IDV to store a `next` value from query parameters, which is used to determine where to redirect the user after completing IDV (or failing to enable their camera).

This also fixes an issue where the course ID was being passed in after accessing IDV from a course, but the user wasn't actually being redirected back to said course.

### DIFF
--- a/src/account-settings/name-change/NameChange.jsx
+++ b/src/account-settings/name-change/NameChange.jsx
@@ -69,7 +69,7 @@ function NameChangeModal({
   useEffect(() => {
     if (saveState === 'complete') {
       handleClose();
-      push('/id-verification');
+      push('/id-verification?next=account%2Fsettings');
     }
   }, [saveState]);
 
--- a/src/id-verification/panels/SubmittedPanel.jsx
+++ b/src/id-verification/panels/SubmittedPanel.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useContext } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { getConfig } from '@edx/frontend-platform';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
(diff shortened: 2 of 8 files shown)


src/account-settings/name-change/NameChange.jsx
```diff
@@ -69,7 +69,7 @@ function NameChangeModal({
   useEffect(() => {
     if (saveState === 'complete') {
       handleClose();
-      push('/id-verification');
+      push('/id-verification?next=account%2Fsettings');
```
MichaelRoytman, line 72: Could you use `encodeURIComponent` here instead? I don't think we should hardcode the URL encoded slash.

src/id-verification/tests/IdVerificationPage.test.jsx
```diff
@@ -41,22 +41,8 @@ describe('IdVerificationPage', () => {
...
-        </IntlProvider>
-      </Router>
-    )));
-    expect(sessionStorage.setItem).toHaveBeenCalledTimes(0);
-  });
-
-  it('does not store empty course_id', async () => {
-    history.push('/?course_id=');
+  it('decodes and stores course_id', async () => {
+    history.push('/?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course');
```
MichaelRoytman, line 45: I think this would be easier to read if you were to use `encoreURIComponent` here.

src/id-verification/panels/SubmittedPanel.jsx
```diff
@@ -19,6 +21,18 @@ function SubmittedPanel(props) {
     });
   }, [userId]);
 
+  // Link back to the correct location if the user accessed IDV somewhere other
+  // than the dashboard
+  useEffect(() => {
+    if (sessionStorage.getItem('courseId')) {
```
MichaelRoytman, line 27: This useEffect hook is duplicated between this panel and the RequestCameraAccessPanel. I wonder if you could write a custom hook or a utility function to avoid the duplication.

PR #570, opened 2022-01-21
[MST-1075] Prevent focus jump when taking photo

bseverino:
https://openedx.atlassian.net/browse/MST-1075

Before this fix, focus would return to the heading when the user takes an IDV photo, causing the scrollbar to jump to the top of the page. The focus should remain on the photo capture button.

### DIFF
--- a/src/id-verification/panels/TakeIdPhotoPanel.jsx
+++ b/src/id-verification/panels/TakeIdPhotoPanel.jsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


bseverino:
So this fix _works_, but I'm still unclear on the cause of the bug in the first place.

According to this line, the header should only be focused when the `BasePanel` mounts, and then only whenever `headingRef.current` is updated (to account for when the heading text changes): https://github.com/openedx/frontend-app-account/blob/32cc2c78350e718dab6baed0c8454fc3944131b9/src/id-verification/panels/BasePanel.jsx#L19

However after some debugging, I confirmed that `headingRef.current` does not update when clicking the photo capture button, which implies that the `BasePanel` is re-rendering for some other reason. I haven't been able to identify where it's coming from.

schenedx (approved):
👍

PR #573, opened 2022-01-25
[MST-869] Remove unused IDV experiment paths

bseverino:
https://openedx.atlassian.net/browse/MST-869

This removes code related to the A/B test to compare camera vs. upload. In the final iteration of this flow:
- The user is required to have a working camera
- Camera is the only option for the portrait photo
- The default mode for ID photo is upload, but the user has the option to switch to camera mode
- The user is given the ability to upload an ID photo on the final step


schenedx (approved):
👍

PR #579, opened 2022-02-15
fix: remove alert and update language if verified name comes from proctored exam attempt

alangsto:
## [MST-1373](https://openedx.atlassian.net/browse/MST-1373)

If a verified name attempt is created by proctoring, we should 1) not show an alert to the user and 2) update the help text for the full name and verified name fields to clarify why the fields are being updated. Maddie from support gave me the copy changes for the help text.

I've included the four scenarios where language is updated in the help text. The language is dependent on the status of the verified name, if the name was created by proctoring, and if the name has been chosen to appear on a certificate.

**Status is submitted (through proctoring), full name should appear on certs:**
Screen Shot 2022-02-15 at 4 40 18 PM

**Status is submitted (through proctoring), verified name should appear on certs:**
Screen Shot 2022-02-15 at 4 39 24 PM

**Status is verified (through proctoring), verified name should appear on certs:**
Screen Shot 2022-02-15 at 4 38 41 PM

**Status is verified (through proctoring), full name should appear on certs:**
Screen Shot 2022-02-15 at 4 38 25 PM

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -259,25 +259,27 @@ class AccountSettingsPage extends React.Component {
     );
   }
 
-  renderFullNameHelpText = (status) => {
-    if (
-      !this.props.verifiedNameHistory
-    ) {
+  renderFullNameHelpText = (status, proctoredExamId) => {
+    if (!this.props.verifiedNameHistory) {
       return this.props.intl.formatMessage(messages['account.settings.field.full.name.help.text']);
     }
 
-    switch (status) {
-      case 'submitted':
-        if (this.props.committedValues.useVerifiedNameForCerts) {
-          return this.props.intl.formatMessage(messages['account.settings.field.full.name.help.text.submitted']);
-        }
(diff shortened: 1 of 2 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -259,25 +259,29 @@ class AccountSettingsPage extends React.Component {
     );
   }
 
-  renderFullNameHelpText = (status) => {
+  renderFullNameHelpText = (status, proctoredExamId) => {
     if (
       !this.props.verifiedNameHistory
```
schenedx, line 264: #NIT I know this is not your code, but this style is super weird. Why can't we have brackets on the same line as the conditional?

schenedx (approved):
The NIT is certainly optional feedback.
LGTM

src/account-settings/AccountSettingsPage.jsx
alangsto: Agreed that this formatting is odd, I'll update it

PR #612, opened 2022-07-19
Paragon form component deprecations

abdullahwaheed:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102#)

### What has changed
Updated deprecated `ValidationFormGroup` to `Form.Group`, `Input` to `Form.Control` and `Checkbox` to `Form. CheckBox` and updated respective unit tests
Created a separate `EditableSelectField` for selections as `EditableField` would become much conditional.


src/account-settings/EditableSelectField.jsx
```diff
@@ -0,0 +1,238 @@
...
+    onChange,
+    isEditing,
+    isEditable,
+    isGrayedOut,
+    intl,
+    ...others
+  } = props;
+  const id = `field-${name}`;
+  let inputOptions = options;
+  if (getConfig().ENABLE_COPPA_COMPLIANCE && name === 'level_of_education' && options) {
```
davidjoy, line 48: Hm, because this conditional is specific to level_of_education, it'd be good to extract it from `EditableSelectField` and modify the `options` as they're passed into the component via props.  That way there's an obvious filtering of options for `level_of_education` where it's being used, rather than quietly changing it within this component where someone might not expect that to happen.

davidjoy (changes requested):
One small-ish factoring change... I also want to pull the branch down and test it.  Were you able to visually verify all these changes worked?  I know some of these UIs are hard to find.

abdullahwaheed:
> One small-ish factoring change... I also want to pull the branch down and test it. Were you able to visually verify all these changes worked? I know some of these UIs are hard to find.

Yes I have tested these changes. Let me share screenshots of changes with PR

davidjoy (approved):

BilalQamar95 (approved):

ishahroz (approved):

PR #623, opened 2022-08-16
feat: add DOB month and year field

alangsto:
If a user is under 13, we would like them to re-enter their DOB for compliance reasons. New DOB fields (for month and year) have been added, and upon submission, will be added to a user's extended profile (as a json object). 

**Concerns:**
There is a UI issue when the backend returns a 400 response related to non-field errors, where the user has to refresh the page in order to submit again, as opposed to just being able to click the submit button again. I noticed this is an issue for all fields on the account settings page, not just the new DOB form.

**UI Elements:** 
Screen Shot 2022-08-18 at 8 47 30 AM
Screen Shot 2022-08-17 at 6 19 44 PM
Screen Shot 2022-08-17 at 6 19 32 PM

**Local testing:** 
Because there aren't unit tests for the account settings page, I tested a few scenarios locally. 

The backend returns a 400 response: 
* expect that save button has an x
* expect error message to show
* local storage key has not been set
* no DOB data has been added to the extended profile

The backend returns a 200 response: 
* expect that save button shows check mark
* expect that component does not render after showing check mark
* local storage key is set
* DOB data is added to extended profile


src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+    return null;
+  }
+
+  useEffect(() => {
+    if (saveState === 'complete') {
+      // set time out so that user can see feedback
+      const timer = setTimeout(() => handleComplete(), 3000);
+      return () => clearTimeout(timer);
+    }
+    return () => null;
```
alangsto, line 54: react complains if there are not consistent returns. This doesn't really mean anything

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+  }, [handleComplete, saveState]);
+
+  return (
+    <form onSubmit={handleSubmit} className="form-group">
```
alangsto, line 58: added in some custom styling to make it look more like the rest of the page

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const month = e.target.month.value;
+    const year = e.target.year.value;
+    const data = month !== '' && year !== '' ? [{ field_name: 'DOB', field_value: `${year}-${month}` }] : [];
+    onSubmit('extended_profile', data);
+  };
+
+  const handleComplete = useCallback(() => {
+    localStorage.setItem('submittedDOB', 'true');
+    dispatch(saveSettingsReset());
```
alangsto, line 34: will cause the page to re-render, and the component will then go away because the local storage key is set

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+    return null;
+  }
+
+  useEffect(() => {
+    if (saveState === 'complete') {
+      // set time out so that user can see feedback
+      const timer = setTimeout(() => handleComplete(), 3000);
+      return () => clearTimeout(timer);
+    }
+    return null;
```
alangsto, line 54: es lint complains if we don't return something here, and returning null doesn't have an impact

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,108 @@
...
+    e.preventDefault();
+
+    const month = e.target.month.value;
+    const year = e.target.year.value;
+    const data = month !== '' && year !== '' ? [{ field_name: 'DOB', field_value: `${year}-${month}` }] : [];
+    onSubmit('extended_profile', data);
+  };
+
+  const handleComplete = useCallback(() => {
+    localStorage.setItem('submittedDOB', 'true');
```
alangsto, line 33: Using local storage instead of a cookie, but let me know if a cookie would be more appropriate. I figured that because this is only being used client side, local storage would be faster.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -494,6 +495,13 @@ class AccountSettingsPage extends React.Component {
     );
 
     const hasLinkedTPA = findIndex(this.props.tpaProviders, provider => provider.connected) >= 0;
+
+    // if user is under 13 and does not have cookie set
+    const shouldUpdateDOB = (
+      getConfig().ENABLE_COPPA_COMPLIANCE
+      && (this.props.formValues.year_of_birth.toString() >= '2009' || this.props.formValues.year_of_birth.toString() === '')
```
alangsto, line 502: This will show if the user is either 13 or has not set a birthdate. Should this be limited to only under 13?

alangsto: Just to clarify, do you mean the value of the year being used for COPPA compliance (aka 2009 here)? I can definitely look into creating a config value for that

I'll also confirm with product about showing the form to users who do not have `year_of_birth` specified.

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -494,6 +495,13 @@ class AccountSettingsPage extends React.Component {
     );
 
     const hasLinkedTPA = findIndex(this.props.tpaProviders, provider => provider.connected) >= 0;
+
+    // if user is under 13 and does not have cookie set
+    const shouldUpdateDOB = (
+      getConfig().ENABLE_COPPA_COMPLIANCE
```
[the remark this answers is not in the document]
alangsto: Yes, it's defined here: https://github.com/edx/edx-internal/blob/943bde9406a41f2b31b6fcf34ad9624eeae8013d/frontends/frontend-app-account/prod_config.yml#L19

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -494,6 +495,13 @@ class AccountSettingsPage extends React.Component {
     );
 
     const hasLinkedTPA = findIndex(this.props.tpaProviders, provider => provider.connected) >= 0;
+
+    // if user is under 13 and does not have cookie set
+    const shouldUpdateDOB = (
```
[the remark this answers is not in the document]
alangsto: Will follow up with product!

src/account-settings/AccountSettingsPage.jsx
alangsto: Confirmed with Spencer that we do not need to show the form to users with no `year_of_birth`, so I will update that

src/account-settings/AccountSettingsPage.jsx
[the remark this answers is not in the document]
alangsto: Spencer confirmed that it is fine to show this form to users if they have not seen the DOB email, so I don't think the querystring param is necessary.

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,147 @@
...
+        <Form.Control.Feedback type="invalid" key="general-error">
+          {intl.formatMessage(messages['account.settingsfield.dob.error.general'])}
+        </Form.Control.Feedback>
+      );
+    }
+    return null;
+  }
+
+  useEffect(() => {
+    if (saveState === 'complete' && isOpen) {
```
alangsto, line 58: have to check that form is open, otherwise other fields will trigger this, as the saveState is controlled by a shared selector

src/account-settings/AccountSettingsPage.jsx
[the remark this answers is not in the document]
alangsto: Product actually wants this behind a flag, so for now I have it behind a config value.

src/account-settings/data/constants.js
```diff
@@ -10,6 +10,11 @@ export const YEAR_OF_BIRTH_OPTIONS = (() => {
   return years.reverse();
 })();
 
+export const COPPA_COMPLIANCE_YEAR = (() => {
```
alangsto, line 13: Instead of adding the year to a config value, I created a constant to decide what year is the cut off

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -494,13 +496,38 @@ class AccountSettingsPage extends React.Component {
...
+      getConfig().ENABLE_COPPA_COMPLIANCE
+      && getConfig().ENABLE_DOB_UPDATE
+      && this.props.formValues.year_of_birth.toString() >= COPPA_COMPLIANCE_YEAR.toString()
+      && localStorage.getItem('submittedDOB') !== 'true'
```
ashultz0, line 505: could this just be localStorage.getItem('submittedDOB')? We're never going to store a no.

src/account-settings/DOBForm.jsx
ashultz0: makes sense to me

src/account-settings/DOBForm.jsx
```diff
@@ -0,0 +1,147 @@
...
+  // eslint-disable-next-line no-unused-vars
+  const [isOpen, open, close, toggle] = useToggle(true, {});
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    const month = e.target.month.value;
+    const year = e.target.year.value;
+    const data = month !== '' && year !== '' ? [{ field_name: 'DOB', field_value: `${year}-${month}` }] : [];
+    onSubmit('extended_profile', data);
```
ashultz0, line 32: is extended profile smart enough not to wipe out anything that is already in there? I'm trying to find the actual call that gets made but still digging

alangsto: It will only wipe out a field if it already exists, so in this case, if there is already a DOB value there, it will write a new one. But all other fields will be left intact - https://github.com/openedx/edx-platform/blob/142945c93026587b6d7f3c36dbbfaf554db8f007/openedx/core/djangoapps/user_api/accounts/api.py#L338-L347

ashultz0: assuming this comes in via the patch (which it would probably have to because there is no post available) _update_extended_profile_if_needed does the right thing

ashultz0 (approved):
not particularly JS proficient but looks like it does what it says :)

schenedx (approved):
👍

PR #613, opened 2022-07-19
Upgraded frontend-build version to v12

BilalQamar95:
**Ticket:**
[42: Upgrade eslint to v8.x](https://github.com/openedx/frontend-wg/issues/42)

**What changed?**
- Updated `frontend-build `to v12 (Eslint was updated in `frontend-build` version resulting in it's version being bumped to v12. This PR updates `frontend-build` to reciprocate eslint version update)
- Resolved `eslint` issues

PR #624, opened 2022-08-18
Revert "Upgraded frontend-build version to v12"

alangsto:
Reverts openedx/frontend-app-account#613

The PR being reverted is causing edits to form fields on the account settings page to error:

Screen Shot 2022-08-18 at 4 29 32 PM (1)

Reverting because we would like to release a form for COPPA compliance, but we are blocked from deploying to prod by the errors from this PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,3 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
 const { createConfig } = require('@edx/frontend-build');
 
 module.exports = createConfig('eslint');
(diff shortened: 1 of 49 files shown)


alangsto:
fyi @BilalQamar95

ashultz0 (approved):

abdullahwaheed:
@alangsto `ValidationFormGroup` is removed in #612 and i have tested is locally. I think we should merge it and this issue will automatically resolve
image

schenedx (changes requested):
This is a huge PR to revert.
I am wondering if we can fix forward.
Do we know the root cause of the bug where the edit name button leads directly to error view?

PR #625, opened 2022-08-19
Reverted changes made to satisfy jsx-no-useless-fragment rule

BilalQamar95:
**Description:** 
Removing fragments to satisfy `react/jsx-no-useless-fragment` resulted in TypeError, reverted those changes & ignored the rule.
Screen Shot 2022-08-19 at 5 22 53 PM


alangsto (approved):
Thanks for the fix! I also tested this locally and everything looks great 👍

PR #630, opened 2022-08-23
perf: preload MFE Config API response in index.html

kdmccormick:
Recently, a capability was added to this MFE that allows it
to load configuration from the LMS's MFE Config
API instead of requiring it to be baked-in to the webpack
build.

When enabled, though, this capability blocks any render
of the MFE until the Config API responds. So, in order to
minimize the amount of time that the MFE is waiting
on the Config API response, we preload the API
response via a <link> in the <head> section
of index.html, which recommends to the browser
to immediately begin loading the MFE Config API
response as soon as index.html is loaded. Otherwise,
the Config API wouldn't be called until React and
frontend-platform are loaded.


kdmccormick:
Closed in favor of #631

PR #641, opened 2022-09-05
Supported Transifex languages in Makefile

abdullahwaheed:
### Ticket
[Expand list of supported Transifex languages to push/pull in Makefile](https://github.com/openedx/frontend-wg/issues/119)

### What has changed
Updated list of Transifex languages in make file to be consistent with Translation Working Groups's active languages. Removed unused files and added new files


dyudyunov:
> @adamstankiewicz i have updated languages list as suggested but in account page, listed languages still differs from the changed languages. Should we update them as well? image

Hi

I'm researching if the list of available languages should show only options added to the dark lang config (as it was on the legacy account page).

I would love to know openedx's plans for it

BilalQamar95 (approved):

ishahroz (approved):
Looks good to me.

PR #702, opened 2022-12-02
fix: prevent users from submitting DOB without update

alangsto:
# [MST-1722](https://2u-internal.atlassian.net/browse/MST-1722)

Users should be forced to update the DOB form before being allowed to submit. Submission is blocked unless the user picks a valid month and year from each drop down, even in the case where the user re-selects the default "Select month" or "Select year" options.

Screen Shot 2022-12-02 at 2 28 26 PM

Screen Shot 2022-12-02 at 2 28 37 PM

Screen Shot 2022-12-02 at 2 28 44 PM

### DIFF
--- a/src/account-settings/DOBForm.jsx
+++ b/src/account-settings/DOBForm.jsx
@@ -3,7 +3,7 @@ import { injectIntl, intlShape } from '@edx/frontend-platform/i18n';
 import {
   Form, StatefulButton, ModalDialog, ActionRow, useToggle, Button,
 } from '@edx/paragon';
-import React, { useCallback, useEffect } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { connect, useDispatch } from 'react-redux';
 import messages from './AccountSettingsPage.messages';
 import { YEAR_OF_BIRTH_OPTIONS } from './data/constants';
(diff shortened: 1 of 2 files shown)


src/account-settings/DOBForm.jsx
```diff
@@ -22,13 +22,23 @@ const DOBModal = (props) => {
...
+    } else if (e.target.name === 'year') {
+      setYearValue(e.target.value);
+    }
+  };
 
   const handleSubmit = (e) => {
     e.preventDefault();
 
-    const month = e.target.month.value;
-    const year = e.target.year.value;
```
alangsto, line 30: no longer need to get them from the event target, as the state is updated each time the value in the form is updated

src/account-settings/DOBForm.jsx
```diff
@@ -22,13 +22,23 @@ const DOBModal = (props) => {
 
   // eslint-disable-next-line no-unused-vars
   const [isOpen, open, close, toggle] = useToggle(true, {});
+  const [monthValue, setMonthValue] = useState('');
+  const [yearValue, setYearValue] = useState('');
+
+  const handleChange = (e) => {
```
alangsto, line 28: there's not a straightforward way to add an event listener to the entire form in react, so choosing to update state based on each form control change.

ashultz0 (approved):

varshamenon4 (approved):

Issue #701, opened 2022-11-29
Page crashes when dark lang is set to partially supported language


timmc-edx:
We've seen bursts of JS errors in the Accounts MFE relating to Beta Languages/dark_lang, specifically in this line:

https://github.com/openedx/frontend-app-account/blob/0e7d2c048d6f36b8f5de9a07ee581224a9aa2022/src/account-settings/BetaLanguageBanner.jsx#L52

This is correlated to traffic from certain countries, and I was able to produce a page crash (although not that specific error) by setting my language to `de` in https://courses.edx.org/update_lang/ -- this causes the page to blank and spew errors in a tight loop.

Post resolution:
- [ ] Restore New Relic alert policy. [See comment](https://github.com/openedx/frontend-app-account/issues/701#issuecomment-1332507288).

ishahroz:
This issue is caused by absence of beta language / dark_lang object literal in `siteLanguageList` at `src -> accout-settings -> site-language -> constants.js`. 

Whenever we set `de` at  https://courses.edx.org/update_lang/ and load the Accounts MFE page, it fetches `de` as `dark-lang` parameter in API response. Later, when `BetaLanguageBanner.jsx` component tries to render, it looks for / filters `de` object literal in `siteLanguageList` via redux-selectors, but finds no such data and just assigns `undefined` to `savedLanguage`. As a result, the code breaks at the line as mentioned in the above comment (no `released` property in undefined object).

For more clarification, I added `de` language literal in `siteLanguageList` and the component started rendering fine, with no page crash errors. Now, as of initial work, **we need to add language literals to support beta languages in Accounts MFE**. In the last, I've a question too. What do we mean by partially supported language? It means MFE translations for the beta languages are in progress? Or is it mean specific language support is only available for certain MFEs?

jmbowman:
Could this have been a side effect of https://github.com/openedx/frontend-app-account/pull/641 , which updated the language list in the Makefile but not in `constants.js`?  If so, we should update the latter to match and also check any other repos where the language list in the Makefile was updated to see if similar changes are needed there.

robrap:
From @timmc-edx:
> (Alert policy https://onenr.io/02R5NWbvkjb is currently set to 30 minutes for 20% error rate, and should probably be restored to 5 minutes.)

I'm guess we'll want to do that once this has been closed.

ishahroz:
[...]

PR #708, opened 2022-12-06
fix: added language codes to support affected languages

ishahroz:
**Ticket**

[Learner Running into "Unexpected Error" when he tried to access Account Settings Page](https://2u-internal.atlassian.net/browse/BOM-3487)

**Related Issue**

[Page crashes when dark lang is set to partially supported language](https://github.com/openedx/frontend-app-account/issues/701)

### DIFF
--- a/src/account-settings/BetaLanguageBanner.jsx
+++ b/src/account-settings/BetaLanguageBanner.jsx
@@ -49,6 +49,9 @@ class BetaLanguageBanner extends React.Component {
 
   render() {
     const savedLanguage = this.getSiteLanguageEntry(this.context.locale);
+    if (!savedLanguage) {
+      return null;
+    }
     const isSavedLanguageReleased = savedLanguage.released === true;
     const noPreviousLanguageSet = this.props.siteLanguage.previousValue === null;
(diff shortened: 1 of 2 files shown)


Mashal-m (approved):
Look good!

BilalQamar95 (approved):

timmc-edx:
I _think_ some of those `released` values are incorrect, but it would be better to merge with incorrect values than to have missing values that cause the page to break. See more info here: https://github.com/openedx/frontend-app-account/issues/701#issuecomment-1352295555 (includes thoughts on a more complete fix that can come afterwards)

robrap:
Related to Tim's comment, don't block yourself on any additional review. Read Tim's linked comment, make the best temporary call on values, and iterate with additional fixes as-needed.

abdullahwaheed (approved):

ishahroz:
In response to Tim's [suggestion](https://github.com/openedx/frontend-app-account/issues/701#issuecomment-1352304731), I've also added a temporary fix to handle unknown languages, so that page maybe might not break for now.

PR #870, opened 2023-09-05
Revert "fix(deps): update dependency @edx/frontend-platform to v5.3.0"

abdullahwaheed:
Reverts openedx/frontend-app-account#867


Syed-Ali-Abbas-Zaidi (approved):

arbrandes:
@Syed-Ali-Abbas-Zaidi, @abdullahwaheed, I didn't catch this locally, apologies.  Let's figure out the best strategy: react-redux 8 is not supposed to be a breaking change, so if I'm reading this right we "just" have to bump the peer deps.

PR #901, opened 2023-10-09
fix: fixed preferences toggle unpredictable behavior 

sundasnoreen12:
[INF-1086](https://2u-internal.atlassian.net/browse/INF-1086)

**Description**

Preferences toggles sometimes have a weird behaviour. 

1. I turned off both toggles the questions toggle turned back on after a few seconds.
2. When it turned it off, the discussions toggle turned back on. 

Solution: Disabled all buttons until the api patch request completes.


awais-ansari (approved):

AhtishamShahid (approved):
it would be nice if there is a unit test for this change.

muhammadadeeltajamul (approved):

PR #912, opened 2023-10-16
refactor: removed disable eslint and created channels constant

awais-ansari:
**[INF-1120](https://2u-internal.atlassian.net/browse/INF-1120)**

- Removed extra spaces from notification preferences module messages.
- Removed disable eslint and refactored code accordingly.
- Create a constant for notification channels and use it accordingly.

### DIFF
--- a/src/notification-preferences/data/constants.js
+++ b/src/notification-preferences/data/constants.js
@@ -1,0 +1,3 @@
+const NOTIFICATION_CHANNELS = ['web'];
+
+export default NOTIFICATION_CHANNELS;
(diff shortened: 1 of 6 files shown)


sundasnoreen12 (approved):

saadyousafarbi (approved):

src/notification-preferences/data/constants.js
```diff
@@ -0,0 +1,3 @@
+const NOTIFICATION_CHANNELS = ['web'];
+
+export default NOTIFICATION_CHANNELS;
```
abdullahwaheed, line 3: since this file will hold multiple constants, i think we should not export this constant as default
```suggestion
export const NOTIFICATION_CHANNELS = ['web'];
```

awais-ansari: We have just this constant right now and maybe in the future, it will remain the same. Apart from that eslint prefers to single export as default. If I make it a name export then I need to disable eslint which is not recommended.

abdullahwaheed (approved):

PR #998, opened 2024-02-12
feat: hide username from account settings page

attiyaIshaque:
### Description
Username visibility on the Account settings page is determined by the backend settings feature flag `ENABLE_AUTO_GENERATED_USERNAME`.

### Screenshots
Before|After|
|-------|-----|
|Screenshot 2024-02-12 at 4 33 11 PM|Screenshot 2024-02-12 at 4 32 44 PM|


syedsajjadkazmii (commented):
[Need opinion] Can we take an approach like, if `ENABLE_AUTO_GENERATED_USERNAME` is enabled on backend, don't pass username key in the API response. On frontend, if we don't have username key in `formValues`, we wont show the username field.

Any cons of this approach? 
cc: @zainab-amir

syedsajjadkazmii:
> [Need opinion] Can we take an approach like, if `ENABLE_AUTO_GENERATED_USERNAME` is enabled on backend, don't pass username key in the API response. On frontend, if we don't have username key in `formValues`, we wont show the username field.
> 
> Any cons of this approach? cc: @zainab-amir

As i think of this now, this approach makes more sense. Thankyou

zainab-amir (approved):

PR #1043, opened 2024-05-08
fix: fixed hover color of cadence button

ayesha-waris:
[INF-1227](https://2u-internal.atlassian.net/browse/INF-1277)
[INF-1360](https://2u-internal.atlassian.net/browse/INF-1360)

- fixed hover colors of email cadence button
- added environment variable for email channel

We now will have a new email setting related to email cadences.
Options for email cadence for a notification can be Daily, Weekly, Immediately and never.
Complete UI as shown in [mockup](https://www.figma.com/file/1WsIDaf9Alfy2r8eNYj6vX/Notifications?type=design&node-id=2072-8529&mode=design&t=2oADgBEXGpGyVx5L-0)


muhammadadeeltajamul (approved):
nit: In `src/index.jsx`, default value for `SHOW_EMAIL_CHANNEL` is of type `boolean` whereas the value of the same environment variable in `.env` file is `string` type. It is better to use same datatype for default value (in this case `boolean`)

src/notification-preferences/NotificationPreferenceColumn.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
+import { getConfig } from '@edx/frontend-platform';
```
sundasnoreen12, line 3: place this import statement to its related group

src/notification-preferences/NotificationPreferenceColumn.jsx
```diff
@@ -82,24 +84,25 @@ const NotificationPreferenceColumn = ({ appId, channel, appPreference }) => {
...
-          'pl-0': channel === 'web' && mobileView,
-        })}
-      >
-        {intl.formatMessage(messages.notificationChannel, { text: channel })}
-      </NavItem>
-      {appPreference
-        ? renderPreference(appPreference)
-        : appPreferences.map((preference) => (renderPreference(preference)))}
-    </div>
+  return ((showEmailChannel === 'true' || channel === 'web') && (
```
sundasnoreen12, line 87: if we have more then 2 channels and showEmailChannel is false then this block will not execute for other channels, i think we need to update this check try below condition instead of web . 
channel !='email'

src/notification-preferences/NotificationPreferenceColumn.jsx
```diff
@@ -82,24 +84,25 @@ const NotificationPreferenceColumn = ({ appId, channel, appPreference }) => {
...
+  <div className={classNames('d-flex flex-column', { 'border-right': channel !== 'email' && showEmailChannel === 'true' })}>
```
sundasnoreen12, line 88: i think the border will be  on the left side if we have more than 1 channels. try and check its behavior. 
above will not work as expected if we will have more than 2 channels

sundasnoreen12 (changes requested):

src/notification-preferences/NotificationPreferenceColumn.jsx
awais-ansari, line 87: `flag || (channel !== "email" && !flag)` This condition will render all the channels regardless of the flag and only render the email channel when the flag is true.

Please move this logic into a helper function.

awais-ansari (changes requested):

src/notification-preferences/data/constants.js
```diff
@@ -1,7 +1,10 @@
...
-  WEB: 'web',
-  EMAIL: 'email',
-};
+import { getConfig } from '@edx/frontend-platform';
+
+const showEmailChannel = getConfig().SHOW_EMAIL_CHANNEL === 'true' || false;
+
+export const NOTIFICATION_CHANNELS = showEmailChannel
+  ? { WEB: 'web', EMAIL: 'email' }
+  : { WEB: 'web' };
```
awais-ansari, line 7: Create a util function for this.

awais-ansari (changes requested):

sundasnoreen12 (changes requested):

sundasnoreen12 (approved):

src/index.scss
```diff
@@ -113,6 +109,15 @@ $fa-font-path: "~font-awesome/fonts";
     font-weight: 500;
   }
 
+  .channel-column:last-child {
+    border-right: 0px !important;
+  }
+
+  .dropdown-item:active,
+  .dropdown-item:focus {
+    background-color: #F2F0EF !important;
```
awais-ansari, line 118: Please use a paragon token instead of a hard-coded color code. like $light-300 == #F2F0EF

awais-ansari (approved):

PR #1044, opened 2024-05-08
feat: use frontend-plugin-framework to provide a FooterSlot

brian-smith-tcril:
* Requires https://github.com/openedx/frontend-plugin-framework/pull/60
* Requires https://github.com/openedx/frontend-plugin-framework/pull/61
* Requires https://github.com/openedx/frontend-component-footer/pull/423

This adds a plugin slot to provide a way to replace the footer using `frontend-plugin-framework`.


arbrandes (approved):
I know this is still a draft, but for what it's worth, I think this is the way to go.  No objections!

arbrandes:
We missed the Redwood cutoff but exactly 60 seconds. 🤷🏼‍♂️

PR #1050, opened 2024-05-16
fix: use `frontend-slot-footer` that supports all footer versions

brian-smith-tcril:
The previous version of `frontend-slot-footer` had a peer dependency of `^14.0.0`, which caused problems for some methods of installing forked footers. This updates to a version of `frontend-slot-footer` that allows for *any* version of `frontend-component-footer` in the peer dependency.


arbrandes (approved):
+1

PR #1063, opened 2024-06-13
build: Update codecov and use token

arbrandes:
Update codecov to the latest version and start using the org-wide token for uploads.

See https://github.com/openedx/wg-frontend/issues/179

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
     - run: make requirements
     - run: make test NPM_TESTS=build
     - run: make test NPM_TESTS=${{ matrix.npm-test }}
-    - name: upload coverage
-      uses: codecov/codecov-action@v3
+    - name: Coverage
+      if: matrix.npm-test == 'test'
+      uses: codecov/codecov-action@v4
       with:
-        fail_ci_if_error: false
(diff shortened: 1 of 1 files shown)


brian-smith-tcril (approved):
Considering this only touches the workflow, it's safe to assume the project coverage issue is unrelated to this PR.

One interesting thing I noticed in the Codecov report here https://github.com/openedx/frontend-app-account/pull/1063/checks?check_run_id=26208113789 was

> Report is 9 commits behind head on master.

That makes me think that something in the past 9 commits has dropped the coverage a little.

Either way this LGTM!

.github/workflows/ci.yml
```diff
@@ -23,7 +23,8 @@ jobs:
     - run: make requirements
     - run: make test NPM_TESTS=build
     - run: make test NPM_TESTS=${{ matrix.npm-test }}
-    - name: upload coverage
-      uses: codecov/codecov-action@v3
+    - name: Coverage
```
brian-smith-tcril, line 26: It seems the matrix on this one is making it tricky (which is what I assume setting `fail_ci_if_error` to `false` was working around). We probably want to make this step *only* run for `test`, not `i18n_extract` or `lint`.

arbrandes: Good call.  I tried with an `if`, now.

PR #1077, opened 2024-07-11
feat!: remove 2U demographics collection

justinhynes:
[APER-2823]

This PR removes functionality specific to 2U and edx.org from the Account MFE that was used for collecting demographics info from learners.


jsnwesson (approved):
Thanks for the cleanup, looks great!

PR #1097, opened 2024-08-09
feat: Added PluginSlot to wrap IDVerificationPage component

rijuma:
### Description

This wraps the IDVerificationPage with a `PluginSlot` component to allow experiment with a new IDV flow.
Similar implementations can be seen in [frontend-app-learner-dashboard](https://github.com/openedx/frontend-app-learner-dashboard/blob/9f52c61e4f5b2098a0253a0ba54cf0ff697dde3b/src/containers/CoursesPanel/index.jsx#L38) for example.

Added a `lint:fix` script as well for practical reasons.

#### How Has This Been Tested?

Tested on local environment and the default behavior is shown as expected.

#### Merge Checklist
* [x] Is there adequate test coverage for your changes?

#### Post-merge Checklist

* [ ] Deploy the changes to prod after verifying on stage or ask **@openedx/edx-infinity** to do it. 
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.


src/index.jsx
```diff
@@ -40,7 +41,14 @@ subscribe(APP_READY, () => {
         >
           <Route path="/notifications/:courseId" element={<NotificationPreferences />} />
           <Route path="/notifications" element={<NotificationCourses />} />
-          <Route path="/id-verification/*" element={<IdVerificationPage />} />
```
schenedx, line 43: With a change like this, if we deploy this to production, the existing IDV flow will be gone. 
I think we need a period of this code where the plugin slot and the existing IDV flow co-exist for a while. This way, while we build out the plugin slot, the existing IDV flow still serves our learners.
Then we need a waffle flag to do the switch from the entry link, which is probably 3 different places.

schenedx (changes requested):
See my inline comment.

src/index.jsx
MichaelRoytman: Whatever component is wrapped by the `PluginSlot` (i.e. `IdVerificationPage`) is the "default component" and will be rendered by default if there is no plugin configuration specified (e.g. in an `env.config.[js|jsx]` file in `edx-internal`) or no overriding plugin in the said plugin configuration, so this deployment strategy isn't necessary. If we merge this as-is, there will be no changes to the IDV page in the learning MFE for 2U or any other deployments. There are more details the [frontend-plugin-framework docs](https://github.com/openedx/frontend-plugin-framework/?tab=readme-ov-file#default-content). Does that assuage your concerns?

MichaelRoytman (commented):
This looks good to me. I know you wanted to discuss more today, so I've held off on approving for now.

Also, should/can we test the use of the `PluginSlot` on the page?

src/index.jsx
rijuma: Indeed. I tested it locally but with no configuration and the `<IdVerificationPage>` is rendered by default.

rijuma:
> This looks good to me. I know you wanted to discuss more today, so I've held off on approving for now.
> 
> Also, should/can we test the use of the `PluginSlot` on the page?

I tested it and the behavior is exactly the same. We could test whatever you have in mind.

MichaelRoytman:
Is it necessary to write a test for the application to ensure that the `PluginSlot` is rendered (e.g something like [this](https://github.com/openedx/frontend-app-learning/blob/99a44dda3773424c221cca04e7a68046dbbbe5a7/src/courseware/course/sequence/Unit/UnitSuspense.test.jsx#L82))?

src/index.jsx
schenedx: Nice. I didn't know plugin slot handles this the way it should be. Thank you for pointing it out.

schenedx (approved):
👍

rijuma:
> Is it necessary to write a test for the application to ensure that the `PluginSlot` is rendered (e.g something like [this](https://github.com/openedx/frontend-app-learning/blob/99a44dda3773424c221cca04e7a68046dbbbe5a7/src/courseware/course/sequence/Unit/UnitSuspense.test.jsx#L82))?

We discussed this offline and we came to the conclusion that adding a test for the index file would be complicated enough to consider it out of scope for this small update.

MaxFrank13:
Hello 👋  I just have one recommendation/request to document the plugin slot using [this format](https://github.com/openedx/frontend-app-account/tree/master/src/plugin-slots) in the MFE. You can use frontend-app-learning for reference as well — [PluginSlots documentation](https://github.com/openedx/frontend-app-learning/tree/99a44dda3773424c221cca04e7a68046dbbbe5a7/src/plugin-slots) & [PluginSlot being implemented ](https://github.com/openedx/frontend-app-learning/blob/99a44dda3773424c221cca04e7a68046dbbbe5a7/src/courseware/course/sequence/Sequence.jsx#L213)

jsnwesson:
Hey all! Agreeing with Max that it would be nice to have a place where all PluginSlots live. You can reference [the Learning MFE](https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots) to see how they organize three PluginSlots. This will also make testing a PluginSlot much easier, I think.

MaxFrank13 (approved):

PR #1116, opened 2024-09-13
feat: added country disabling feature

awais-ansari:
[INF-1567](https://2u-internal.atlassian.net/browse/INF-1567)
### Description

- Countries in the `disabledCountries` list will no longer be visible to users in the account settings of other countries, and users will not be able to switch to disabled countries. 
- User with a disabled country selected can change their country to another one. After switching, they cannot select the disabled country again.
- Implemented frontend validation to verify the country against the disabled countries list to prevent invalid country submissions.

For Example, if Russia is on the disabled countries list.
> Other Country Users will not see Russia in the country list as an 
Screenshot 2024-09-14 at 2 10 05 PM
Screenshot 2024-09-14 at 2 07 52 PM

------------------------------------------------------------------------------------------------------------------------
> Russian user will see Russia in the country list but it'll be disabled and they can change their country to another one.
Screenshot 2024-09-14 at 2 09 07 PM
Screenshot 2024-09-14 at 2 09 30 PM

------------------------------------------------------------------------------------------------------------------------
**Note**
- The disabledCountries list will be fetched from an API in phase 2 implementation.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -120,7 +120,15 @@ class AccountSettingsPage extends React.Component {
     countryOptions: [{
       value: '',
       label: this.props.intl.formatMessage(messages['account.settings.field.country.options.empty']),
-    }].concat(getCountryList(locale).map(({ code, name }) => ({ value: code, label: name }))),
+    }].concat(
+      this.removeDisabledCountries(
+        getCountryList(locale).map(({ code, name }) => ({
+          value: code,
+          label: name,
+          disabled: this.isDisabledCountry(code),
+        })),
+      ),
+    ),
     stateOptions: [{
       value: '',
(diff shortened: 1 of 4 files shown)


AhtishamShahid (approved):

src/account-settings/AccountSettingsPage.jsx
```diff
@@ -147,11 +155,30 @@ class AccountSettingsPage extends React.Component {
...
 
+  removeDisabledCountries = (countryList) => {
+    const { disabledCountries, committedValues } = this.props;
+
+    if (!disabledCountries.length) {
+      return countryList;
+    }
+
+    return countryList.filter(({ value }) => {
+      const isDisabled = this.isDisabledCountry(value);
```
sundasnoreen12, line 166: There's no need to recalculate isDisabled here since we are already receiving disabled as a parameter in the countryList. as countryList.filter(({ value , disabled})

sundasnoreen12 (changes requested):

PR #1260, opened 2025-06-02
refactor: moved hardcoded support urls to constants

ayesha-waris:
[INF-1742](https://2u-internal.atlassian.net/browse/INF-1742)
### Description

The Account MFE has a number of [hardcoded URLs that go to edX.org](https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-account++%22.edx.org%2F%22&type=code). In several cases, these are hardcoded in React components, and they go to 2U  resources which has been moved to constants in this PR.


awais-ansari:
Add these URL in edx-internal and edge-internal

PR #1273, opened 2025-06-18
sync `master` -> `master-design-tokens`

brian-smith-tcril:
# LAND AS MERGE COMMIT

I resolved the package lock conflict by keeping the existing version, then I set the updated dependency versions directly in `package.json` and ran `npm install` to update the package lock.


arbrandes (approved):
👍🏼

PR #1274, opened 2025-06-19
fix: use Container component with size='xl'

adamstankiewicz:
### Description

By default, this MFE does not restrict the max-width of its main content area, resulting in reduced readability due to longer lines of text, especially on wider screens. Most MFEs constrain containers with a max-width for this reason.

#### Local (elm-theme, Paragon v23)

Replaces hardcoded `.container-fluid` with `<Container size="xl">` to ensure the main content area does not stretch the full viewport.

image

#### Stage (elm-theme, Paragon v23)

No max-width, resulting in the main content area expanding full-width which is not ideal for readability (line length) and larger screens.

image

#### Prod (brand-edx.org, Paragon v22)

image

#### Merge Checklist

* [ ] If your update includes visual changes, have they been reviewed by a designer? Send them a link to the Sandbox, if applicable.
* [ ] Is there adequate test coverage for your changes?

#### Post-merge Checklist

* [ ] Deploy the changes to prod after verifying on stage or ask **@openedx/edx-infinity** to do it. 
* [ ] 🎉 🙌 Celebrate! Thanks for your contribution.

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -14,7 +14,7 @@ import {
   getLanguageList,
 } from '@edx/frontend-platform/i18n';
 import {
-  Hyperlink, Icon, Alert,
+  Container, Hyperlink, Icon, Alert,
 } from '@openedx/paragon';
 import { CheckCircle, Error, WarningFilled } from '@openedx/paragon/icons';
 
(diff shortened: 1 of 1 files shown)


src/account-settings/AccountSettingsPage.jsx
```diff
@@ -852,7 +852,7 @@ class AccountSettingsPage extends React.Component {
     } = this.props;
 
     return (
-      <div className="page__account-settings container-fluid py-5">
```
adamstankiewicz, line 855: [inform] the removal of `container-fluid` is intentional; `fluid` is true by default.

brian-smith-tcril (approved):
This seems like a great solution to me!

PR #1328, opened 2025-08-18
Revert "Replace of injectIntl with useIntl() 10/10"

sundasnoreen12:
Reverts openedx/frontend-app-account#1314

Hi @diana-villalvazo-wgu, I’m reverting this PR as it is causing issues in the stage environment. I’ve attached the error details for your reference.

Screenshot 2025-08-18 at 12 55 14 PM

### DIFF
--- a/src/account-settings/AccountSettingsPage.jsx
+++ b/src/account-settings/AccountSettingsPage.jsx
@@ -1,18 +1,17 @@
 import { AppContext } from '@edx/frontend-platform/react';
-import { getConfig } from '@edx/frontend-platform';
-import {
-  useEffect, useContext, useMemo, createRef, useState,
-} from 'react';
+import { getConfig, getQueryParameters } from '@edx/frontend-platform';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import memoize from 'memoize-one';
 import findIndex from 'lodash.findindex';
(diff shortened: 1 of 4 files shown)


deborahgu:
@diana-villalvazo-wgu this is starting to be a real problem,  because right now the repository is broken, and that  is something that has to get fixed -- and tested -- before the Ulmo cut.   Are you able to prioritize this? I see the draft PR you link hasn't been touched in 3 weeks.

@arbrandes, on the assumption that @diana-villalvazo-wgu  isn't available to help us address this, can we disable the code coverage requirement for now so that we can do this reversion? Thanks for your assist.

diana-villalvazo-wgu:
> @diana-villalvazo-wgu this is starting to be a real problem, because right now the repository is broken, and that is something that has to get fixed -- and tested -- before the Ulmo cut. Are you able to prioritize this? I see the draft PR you link hasn't been touched in 3 weeks.
> 
> @arbrandes, on the assumption that @diana-villalvazo-wgu isn't available to help us address this, can we disable the code coverage requirement for now so that we can do this reversion? Thanks for your assist.

I think should be ok to proceed with this revert, as i mentioned the tests seems to be not passing coverage since on my PR i improved it and now this PR is reverting it so it should be ok is expected

diana-villalvazo-wgu:
We proceed to merge this, since we have a known reason why the coverage is going down (reverting a PR) 👍 
cc. @deborahgu / @sundasnoreen12 / @awais-ansari

PR #1377, opened 2025-11-14
fix: remove 2u specific links

MaxFrank13:
This updates hardcoded links to instead use their corresponding configuration values. The hardcoded values were only relevant for 2u/edX


sarina:
Sorry for the late reply. I agree with your proposed defaults. Thanks!

Issue #1466, opened 2026-09-22
Convert to frontend-base


arbrandes:
### Description

Convert the app into a frontend-base App Repository, following https://github.com/openedx/frontend-base/blob/main/docs/how_tos/migrate-frontend-app.md: drop frontend-platform, frontend-build, the header and footer components and the plugin framework in favor of frontend-base, add a site config and app entry point, move header and footer customizations to slots, make the package a buildless library, fix the tests and modernize the README.

This is the second PR in the stack described in #1316, on top of the react-query conversion. The previous attempt, #1325, was closed and this starts over. The existing `frontend-base` branch is an untouched snapshot of an old `master` and can be reset before work starts.

Features that don't survive the conversion (react-helmet, Optimizely, and whatever else turns up) will be listed here along with their DEPR tickets.

Issue #1316, opened 2025-07-31
Convert Account to frontend-base


arbrandes:
### Description

Convert the Account app to use frontend-base, as per https://github.com/openedx/frontend-base/blob/main/docs/how_tos/migrate-frontend-app.md.

The converted code will remain in a long-lived [`frontend-base`](https://github.com/openedx/frontend-app-account/tree/frontend-base) branch.

### Dropped features

The following features are removed as part of the conversion:

- ...

DEPR tickets will be created and linked above as necessary.

### PRs

- https://github.com/openedx/frontend-app-account/pull/1325

### TODO

- [ ] Remove undesired features
- [ ] Convert basic app structure (site.config, App.tsx, etc)
- [ ] Refactor imports 
- [ ] Convert header
- [ ] Convert footer
- [ ] Fix tests
- [ ] Publish to NPM as a buildless library
- [ ] Include in frontend-template-site as default configuration
- [ ] Paragon 23 (depends on frontend-base support)
- [ ] Ongoing: merges with master

brian-smith-tcril:
I would like to take care of:
* https://github.com/openedx/frontend-app-account/pull/1368
* https://github.com/openedx/paragon/issues/3963

before doing this conversion

PR #1468, opened 2026-09-22
refactor: convert from Redux to react-query

### LINKED ISSUE #1465: Convert to react-query
### Description

Replace the Redux stack (redux, react-redux, redux-saga, redux-thunk, reselect and their dev tooling) with @tanstack/react-query, so server state is fetched, cached and mutated through query and mutation hooks instead of reducers and sagas. This covers the store in `src/data`, the account-settings reducers and sagas (including site-language, name-change, third-party-auth, reset-password and delete-account) and notification-preferences.

This is the first PR in the stack described in #1316. Doing it on the current codebase, before the frontend-base conversion, keeps the two changes reviewable separately. openedx/frontend-app-learner-dashboard#786 did the equivalent for Learner Dashboard.

arbrandes:
### Description

Replaces the Redux stack (redux, react-redux, redux-saga, redux-thunk, reselect) with @tanstack/react-query, in two commits: the notification preferences first, then account settings along with the removal of the Redux dependencies. Server state lives in per-resource queries, the shared form state (the open field, drafts, per-field errors, pending confirmations) in a React context with a small reducer, and the class components become functions. Every `service.js` is renamed to `api.js`, following Learner Dashboard.

Behavior changes worth knowing about: an unexpected error during a save, password reset or account deletion used to cancel the root saga and leave the page unresponsive until reload; it is now logged and contained to the request that raised it. A failed notification toggle no longer empties the notifications section. The Reset Password button now really ignores clicks while a request is pending, as its comment always claimed. The `onEdit`/`onCancel` props of the field components are gone, since `connect` was already overriding them.

The `QueryClientProvider` in `index.jsx` is temporary: the frontend-base conversion (#1466) drops it in favor of the shell's. This is the first PR in the stack described in #1316. Closes #1465.

### LLM usage notice

Built with assistance from Claude.


src/account-settings/data/FormContext.jsx
```diff
@@ -0,0 +1,190 @@
...
+      for (let i = 0; i < settingsArray.length; i += 1) {
+        const { formId, commitValues } = settingsArray[i];
+        const commitData = { [formId]: commitValues };
+        dispatch({ type: SAVE_BEGIN });
+        // eslint-disable-next-line no-await-in-loop
+        const savedValues = await patchSettings(username, commitData, userId);
+        applySavedValues(savedValues, commitData);
+      }
+    },
+    onMutate: () => dispatch({ type: SAVE_BEGIN }),
```
jacobo-dominguez-wgu: Is not redundant to have `dispatch({ type: SAVE_BEGIN }),` in here and also on the `mutationFn`?

src/notification-preferences/NotificationPreferences.jsx
```diff
@@ -65,17 +52,6 @@ const NotificationPreferences = () => {
         </div>
       )}
       {preferencesList}
-      {isLoading && (
```
jacobo-dominguez-wgu: Self note: the spinner was removed because it was inseparable from the Redux LOADING_STATUS state that no longer exists, and React Query's loading state wasn't wired up to a replacement indicator — the component just renders null until data arrives, leaning on the page-level loader for the initial-load case.

PR #1469, opened 2026-09-22
feat!: convert to frontend-base

### LINKED ISSUE #1466: Convert to frontend-base

arbrandes:
### Description

Converts the app into a frontend-base App Repository per the migration how-to: frontend-platform, frontend-build, the header and footer components and the plugin framework are replaced by `@openedx/frontend-base` 2.0.0-alpha; the package becomes a buildless library exporting `accountApp` and `accountRoutes`; configuration moves to `site.config` with the app's own keys bundled as `defaultConfig` under their existing names, so MFE config API values keep flowing. The settings page mounts at `account` under the `org.openedx.frontend.role.account` role, ID verification stays at `account/id-verification` for the LMS's links and gets its own role.

The two plugin slots become `Slot`s named per ADR 0009 (`org.openedx.frontend.slot.account.additionalProfileFields.v1` and `org.openedx.frontend.slot.account.idVerificationPage.v1`); the footer slot documentation goes, as the footer is the shell's. The country and language lists come from `i18n-iso-countries` and `@cospired/i18n-iso-languages`, as frontend-platform's did.

Breaking changes for operators: the site language select now offers the languages the site bundles translations for (the shell's list), so the hard-coded list and the beta language banner are gone; the Optimizely script, the Usabilla widget and `LEARNER_FEEDBACK_URL` are dropped along with the legacy `index.html` and document head; the shell handles unknown routes instead of `NotFoundPage`. DEPR tickets for Usabilla and the beta language banner: TBD.

Second PR in the stack described in #1316, on top of #1468. Closes #1466.

### LLM usage notice

Built with assistance from Claude.


arbrandes:
Superseded by #1470, which comes from a branch on this repository so it can be stacked on #1468.

